PR #384, opened 2016-07-01
Limit width for emails

iwz:
When a very large image comes in through email, the CSS `max-width` does
not apply in some email clients (Airmail, Gmail, Google Inbox), so the
email appears extremely wide and you must scroll left/right to read each
paragraph.

This is an attempt to limit the email width.
The max width for the New Announcement and New Comment emails should now
be 600px.

See: http://templates.mailchimp.com/development/responsive-email/fluid-images/


iwz:
@BlakeWilliams @paulcsmith ready for review

BlakeWilliams:
LGTM! Thanks for working on this.

PR #420, opened 2016-10-04
Fix asset compilation

romulomachado:
* Using the :root endpoint configuration for watchers is deprecated. The :cd option at the end of the watcher argument list is now required.
* Update brunch to work with node v6.6.0.


BlakeWilliams:
This looks good, thanks!

PR #424, opened 2016-10-18
Clean up .sample.env

BlakeWilliams:
This removes the unused `HUB_API_TOKEN` and adds `CLIENT_ID` and
`CLIENT_SECRET` to make it more clear that they need to be fetched from
Github.

### DIFF
--- a/.sample.env
+++ b/.sample.env
@@ -2,4 +2,5 @@ MANDRILL_KEY=GetFromMandrillSettings
 OUTBOUND_EMAIL_DOMAIN=staging.constable.io
 INBOUND_EMAIL_DOMAIN=inbound.staging.constable.io
 SLACK_WEBHOOK_URL=https://fakeslack/webhook
-HUB_API_TOKEN=getFromHubIfNeeded
+CLIENT_ID=GetFromGithub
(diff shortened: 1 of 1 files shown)


SViccari:
Perfect! I was just about to issue a PR with this change as I found not having the CLIENT_ID set results in tests failing locally.

.sample.env
```diff
@@ -2,4 +2,5 @@ MANDRILL_KEY=GetFromMandrillSettings
 OUTBOUND_EMAIL_DOMAIN=staging.constable.io
 INBOUND_EMAIL_DOMAIN=inbound.staging.constable.io
 SLACK_WEBHOOK_URL=https://fakeslack/webhook
-HUB_API_TOKEN=getFromHubIfNeeded
+CLIENT_ID=GetFromGithub
+CLIENT_SECRET=GetFromGithub
```
SViccari: Where in Github would one find the correct CLIENT_ID? I'm guessing there is a shared resource that only Thoughtbot individuals can access?

SViccari: ah, okay 😄  thanks!

SViccari:
I think it's time to 🚢  this 😉

Issue #417, opened 2016-10-04
Use the the permitted email domain in the error message when failing to sign up


paulcsmith:
https://github.com/thoughtbot/constable/blob/45e5e8591db15c2ab4adc23acae0f91023bc7a10/web/controllers/auth_controller.ex#L36 this should instead use https://github.com/thoughtbot/constable/blob/45e5e8591db15c2ab4adc23acae0f91023bc7a10/config/config.exs#L27

See #357

mjankowski:
Done here - https://github.com/thoughtbot/constable/commit/ef43c2033cc633affeff92e579835a9796093583

PR #425, opened 2016-10-23
use configured email domain in tests

SViccari:
I was initially going to address Issue #417 but noticed @BlakeWilliams already took care of it :) I noticed that some of the tests aren't using the configured email domain and opted to address that instead.

Why:
Changing the configured, permitted email domain results in 8 test
failures.

This PR:
Uses the configured email domain in tests to prevent unnecessary test
failures should the permitted email domain be altered. This PR does
leave one test that explicitly checks for the thoughtbot email domain to
prevent typos or accidental changes from sneaking in.

### DIFF
--- a/test/controllers/auth_controller_test.exs
+++ b/test/controllers/auth_controller_test.exs
@@ -3,7 +3,9 @@ defmodule AuthControllerTest do
   alias Constable.{User, UserIdentifier, UserInterest}
 
   @google_authorize_url "https://accounts.google.com/o/oauth2/auth"
-  @oauth_email_address "fake@thoughtbot.com"
+  @permitted_email_domain Application.fetch_env!(:constable, :permitted_email_domain)
+ 
+  def valid_email_address, do: "fake@#{@permitted_email_domain}"
 
   defmodule FakeTokenRetriever do
     def get_token!(_conn, _code, _token_params) do
@@ -12,12 +14,10 @@ defmodule AuthControllerTest do
   end
 
   defmodule FakeRequestWithAccessToken do
-    @oauth_email_address "fake@thoughtbot.com"
-
     def get!(_token, _path) do
       %OAuth2.Response{
         body: %{
-          "email" => @oauth_email_address,
+          "email" => AuthControllerTest.valid_email_address,
           "name" => "Gumbo McGee"
         }
       }
@@ -25,12 +25,10 @@ defmodule AuthControllerTest do
   end
 
   defmodule NonThoughtbotRequestWithAccessToken do
-    @oauth_email_address "fake@example.com"
-
     def get!(_token, _path) do
       %OAuth2.Response{
         body: %{
-          "email" => @oauth_email_address,
+          "email" => "fake@not_permitted.com",
           "name" => "Gumbo McGee"
         }
       }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -23,6 +23,7 @@
   "devDependencies": {
     "babel-brunch": "~6.0.0",
     "babel-preset-es2015": "^6.6.0",
+    "babel-preset-es2016": "^6.16.0",
```
SViccari: fwiw, before adding this to the dev dependencies, I also tried the following steps, in hopes it would fix the issue instead:
```
rm -rf ~/.npm
rm -rf node_modules/
npm install
```

MattMSumner: Should this replace `babel-preset-es2015`?

test/controllers/auth_controller_test.exs
```diff
@@ -3,7 +3,9 @@ defmodule AuthControllerTest do
   alias Constable.{User, UserIdentifier, UserInterest}
 
   @google_authorize_url "https://accounts.google.com/o/oauth2/auth"
-  @oauth_email_address "fake@thoughtbot.com"
+  @permitted_email_domain Application.fetch_env!(:constable, :permitted_email_domain)
+ 
+  def valid_email_address, do: "fake@#{@permitted_email_domain}"
```
MattMSumner: Is there a difference between the function definition here vs the variable definition in `user_test` (valid_email "foo@#{@permitted_email_domain}"`)? Is one preferable over the other?

test/models/user_test.exs
```diff
@@ -34,15 +37,15 @@ defmodule Constable.UserTest do
...
   end
 
   test "create_changeset sets name from username only if the name is blank" do
-    changeset = User.create_changeset(%User{}, %{email: "foo@thoughtbot.com", name: "Real Name"})
+    changeset = User.create_changeset(%User{}, %{email: @valid_email, name: "Real Name"})
     assert changeset.changes[:name] == "Real Name"
 
     username = "foobar"
-    changeset = User.create_changeset(%User{}, %{email: username <> "@thoughtbot.com"})
+    changeset = User.create_changeset(%User{}, %{email: username <> "@#{@permitted_email_domain}"})
```
MattMSumner: It looks like we're interpolating and concat-ing here? How would you feel about only using interpolation? i.e. 
```ex
"#{username}@#{@permitted_email_domain}"
```

paulcsmith:
I think this is good to merge once comments from Matt are addressed. Thanks @SViccari

test/controllers/auth_controller_test.exs
SViccari: Great question! There is indeed a reason that this file uses a function over a module attribute. 

If the valid email address were defined using a module attribute, the other modules in this file, `FakeRequestWithAccessToken` and `FakeTokenInfoGoogleStrategy`, don't have access to `@valid_email_address`, as it's defined on `AuthControllerTest`. To still define a valid email address format in one location, the best way around this seems to be to define a function on `AuthControllerTest` and call `AuthControllerTest.valid_email_address` within the other modules.

I'm certainly open to an alternative approach if you have one :)

test/models/user_test.exs
SViccari: Good catch, I'll update!

package.json
SViccari: Oh, good thought. I made the assumption that es2015 might still be in use but perhaps that's not the case.

BlakeWilliams: I don't think so, what makes this dependency necessary?

SViccari:
@MattMSumner Thanks for the review! I pushed up two commits to address your feedback. I'll squash these commits once these changes are approved.

package.json
BlakeWilliams: I'm not positive, but I want to say that you do. I don't think es2016 includes all that es2015 did.

SViccari:
@BlakeWilliams @MattMSumner I believe I've found a better solution to the `couldn't find preset` compile warning for babel-preset-es2016 (which also causes tests to fail locally). I've removed any changes related to babel-preset-es2016 so this PR only alters tests. Since I've separated the concerns, I believe this PR is ready to be merged now :)

I'll issue a fix for the babel-preset-es2016 warning shortly.

BlakeWilliams:
Awesome, this looks good to me. Thanks!

PR #429, opened 2016-10-28
Reformat daily digest text email

kenyonj:
Before:

After:

### DIFF
--- a/web/templates/email/daily_digest.text.eex
+++ b/web/templates/email/daily_digest.text.eex
@@ -1,8 +1,8 @@
-<%= gettext("# Daily digest of new Interests and Announcements") %>
+<%= gettext("# Daily digest of new Interests and Announcements\n") %>
 
 <%= unless Enum.empty?(@interests) do %>
 <%= gettext("Recently added interests") %>
--------------
+-----------------
 
 <%= for interest <- @interests do %>
   #<%= interest.name %> - <%= interest_url(Constable.Endpoint, :show, interest) %>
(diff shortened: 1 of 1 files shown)


web/controllers/email_preview_controller.ex
```diff
@@ -9,11 +9,11 @@ defmodule Constable.EmailPreviewController do
       email_body(email_name) |> Repo.rollback
     end
 
-    html(conn, body)
+    text(conn, body)
   end
 
   defp email_body(email_name) do
-    email_for(String.to_atom(email_name)).html_body
+    email_for(String.to_atom(email_name)).text_body
```
SViccari: Was this intended to be a temporary change? Or do you prefer for the email preview to display the text version vs the html version?

web/templates/email/daily_digest.text.eex
```diff
@@ -1,31 +1,33 @@
...
 <% end %>
 <%= unless Enum.empty?(@announcements) do %>
-<%= gettext("Recently added announcements") %>
+<%= gettext("Recently added announcements\n") %>
 -----------------
 
 <%= for announcement <- @announcements do %>
   <%= announcement.title %> - <%= announcement_url(Constable.Endpoint, :show, announcement) %>
   <%= gettext("posted by") %> <%= announcement.user.name %> <%= gettext("in") %> <%= interest_names(announcement) %>
+  <%= gettext("\n") %>
```
SViccari: Perhaps `<%= "\r\n" %>` would work here as well, and below on [line 30](https://github.com/thoughtbot/constable/pull/429/files#diff-bd64116ecaddf010b9c23b36a38cdee3R30)?

web/controllers/email_preview_controller.ex
kenyonj: Yes, this was temporary, thanks for catching that! Fixed in 693eb62

web/templates/email/daily_digest.text.eex
```diff
@@ -1,31 +1,33 @@
...
 <% end %>
 <%= unless Enum.empty?(@comments) do %>
-<%= ngettext("%{count} announcement was discussed", "%{count} announcements were discussed", length(discussed_announcements(@comments))) %>
+<%= ngettext("%{count} announcement was discussed ", "%{count} announcements were discussed ", length(discussed_announcements(@comments))) %>
 <%= ngettext("by %{count} person", "by %{count} people", length(unique_commenters(@comments))) %>
 
 <%= for announcement <- discussed_announcements(@comments) do %>
   <%= announcement.title %> - <%= announcement_url(Constable.Endpoint, :show, announcement) %>
   <%= new_comment_count_text(@comments, announcement) %> <%= commenters_list_text(announcement, @comments) %>
+  <%= gettext("\n") %>
```
paulcsmith: Same here. I don't think `gettext` is needed

paulcsmith:
Thanks for the review @SViccari, and nice find @kenyonj. LGTM!

web/templates/email/daily_digest.text.eex
kenyonj: agreed, thanks @SViccari and @paulcsmith ... I think we can get away with just the newline character `\n` without the need for the carriage return character `\r`.

Fixed in 0ebf7f3

kenyonj:
Thanks for the review @SViccari and @paulcsmith

PR #432, opened 2016-11-06
Replace deprecated `Dict.fetch!` with `Map.fetch!`

DNNX:
`Dict` module is deprecated since Elixir [1.2](https://github.com/elixir-lang/elixir/blob/v1.2/lib/elixir/lib/dict.ex#L3).
Let's use `Map` instead.


SViccari:
Good catch! :shipit:

paulcsmith:
Thanks @DNNX!

Issue #416, opened 2016-09-27
Better page titles


derekprior:
The page title for all pages seems to be `Constable`. We should use the announcement or tag title when appropriate. I dislike having several `Constable` tabs in my browser.
![constable](https://cloud.githubusercontent.com/assets/152152/18876134/85f7b33a-8496-11e6-9d6c-a96c6342e6c1.png)

PR #433, opened 2016-11-18
Add better page titles

paulcsmith:
Closes: https://github.com/thoughtbot/constable/issues/416


drapergeek:
I like it! 👍

web/controllers/search_controller.ex
```diff
@@ -10,6 +10,7 @@ defmodule Constable.SearchController do
     |> assign(:announcements, announcements.entries)
     |> assign(:search_page, announcements)
     |> assign(:query, params["query"])
+    |> page_title("Results for #{params["query"]}")
```
joshuaclayton: thoughts on single quotes around the search string?

paulcsmith: Good idea!

Issue #435, opened 2017-02-10
Don't search "This week" or "SRS BSNS"


croaky:
I've been finding the experience of https://search.thoughtbot.com has been degrading over time as SRS BSNS increasingly takes up the results of various searches.

For example the "right" answer for https://search.thoughtbot.com/?utf8=%E2%9C%93&q=business+card is buried.

Think one thing we can do in Search is make Constable be last in results, or Handbook first, but another improvement would also be to have Constable search no find SRS BSNS results.

Thoughts? @georgebrock @drapergeek

drapergeek:
I agree, that's frustrating.

I think listing the results last is a quick fix but SRS BSNS encourages more and more duplication.

Maybe there should be some way to tag a constable post as not to include it in search?

georgebrock:
I like the idea of excluding certain posts from the constable results. How about an `exclude_topics` parameter on Constable search? That way, `search.thoughtbot.com` could pass `exclude_topics=serious-business`, but `constable.io/search` wouldn't.

drapergeek:
It seems odd that search would know what is important to exclude and what is not.

croaky:
Some kind of parametrized filters on search makes sense to me, and the Search service can set the default filters that make sense for it.

BlakeWilliams:
I'm a fan of adding an `exclude_interests` field so we can ignore specific interests. I may jump on that today.

croaky:
Thanks for the fixes!

PR #439, opened 2017-02-24
Allow excluding specific interests in search

BlakeWilliams:
This implements the ability to exclude specific interests from returned
search results. This will be helpful to tools that want to search
Constable but exclude specific interests.

This adds the feature to both the API and non-API search controller's
for parity but doesn't expose UI to use this functionality.

### DIFF
--- a/test/controllers/api/search_controller_test.exs
+++ b/test/controllers/api/search_controller_test.exs
@@ -22,8 +22,27 @@ defmodule Constable.Api.SearchesControllerTest do
     assert results_for(conn, query: query_with_extra_spaces) == json_for(announcement_2)
   end
 
+  test "returns matching announcements that don't have the excluded interest", %{conn: conn} do
+    lame = insert(:interest, name: "lame")
+    announcement_1 = insert(:announcement, title: "foobar1") |> tag_with_interest(lame)
+    announcement_2 = insert(:announcement, body: "announcement body cool")
+    announcement_3 = insert(:announcement, title: "awesome title", body: "cool body")
+
+    assert results_for(conn, query: "foobar1", exclude: ["lame"]) == json_for([])
+    assert results_for(conn, query: "announcement body", exclude: ["lame"]) == json_for(announcement_2)
+    assert results_for(conn, query: "awesome title cool body", exclude: []) == json_for(announcement_3)
+  end
+
   defp results_for(conn, query: search_term) do
-    response = post conn, api_search_path(conn, :create), query: search_term
+    results_for(conn, query: search_term, exclude: [])
+  end
+
+  defp results_for(conn, query: search_term, exclude: exclude) do
+    response = post conn,
+      api_search_path(conn, :create),
+      query: search_term,
+      exclude_interests: exclude
+
     json_response(response, 200)
   end
 
--- a/test/controllers/search_controller_test.exs
+++ b/test/controllers/search_controller_test.exs
@@ -1,0 +1,34 @@
+defmodule Constable.SearchControllerTest do
+  use Constable.ConnCase, async: true
+
+  setup do
+    {:ok, browser_authenticate()}
+  end
+
+  test "#show returns announcements matching search", %{conn: conn}do
+    foo = insert(:interest, name: "foo")
+    insert(:announcement, title: "Awesome Post") |> tag_with_interest(foo)
+    insert(:announcement, title: "Not so much post", body: "awesome post!") |> tag_with_interest(foo)
+    insert(:announcement, title: "Don't show up!", body: "lame post!") |> tag_with_interest(insert(:interest))
+
+    conn = get conn, search_path(conn, :show, query: "awesome")
+
+    assert html_response(conn, :ok) =~ "Awesome Post"
+    assert html_response(conn, :ok) =~ "Not so much post"
+    refute html_response(conn, :ok) =~ "Don't show up!"
+  end
+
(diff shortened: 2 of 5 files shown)


BlakeWilliams:
Should help resolve #435

test/controllers/api/search_controller_test.exs
```diff
@@ -22,8 +22,27 @@ defmodule Constable.Api.SearchesControllerTest do
...
+    assert results_for_exclude(conn, query: "announcement body", exclude: ["lame"]) == json_for(announcement_2)
+    assert results_for_exclude(conn, query: "awesome title cool body", exclude: []) == json_for(announcement_3)
+  end
+
   defp results_for(conn, query: search_term) do
-    response = post conn, api_search_path(conn, :create), query: search_term
+    results_for_exclude(conn, query: search_term, exclude: [])
+  end
+
+  defp results_for_exclude(conn, query: search_term, exclude: exclude) do
```
paulcsmith: It's unclear what is being excluded when looking at just the test. Thoughts on changing this to `exclude_interest`?

paulcsmith: And maybe it could use the same function name as above? `results_for(conn, query: "announcement_body", exclude_interests: ["lame"])`?

test/controllers/api/search_controller_test.exs
```diff
@@ -22,8 +22,27 @@ defmodule Constable.Api.SearchesControllerTest do
     assert results_for(conn, query: query_with_extra_spaces) == json_for(announcement_2)
   end
 
+  test "excludes announcements that have an excluded interest", %{conn: conn} do
```
paulcsmith: Maybe "returns matching announcements that don't have the excluded interest"?

test/controllers/search_controller_test.exs
```diff
@@ -0,0 +1,34 @@
...
+    insert(:announcement, title: "Awesome Post") |> tag_with_interest(foo)
+    insert(:announcement, title: "Not so much post", body: "awesome post!") |> tag_with_interest(foo)
+    insert(:announcement, title: "Don't show up!", body: "lame post!") |> tag_with_interest(insert(:interest))
+
+    conn = get conn, search_path(conn, :show, query: "awesome")
+
+    assert html_response(conn, :ok) =~ "Awesome Post"
+    assert html_response(conn, :ok) =~ "Not so much post"
+    refute html_response(conn, :ok) =~ "Don't show up!"
+  end
```
paulcsmith: Nice catch. I'm glad we have a test for this now

web/controllers/api/search_controller.ex
```diff
@@ -5,8 +5,13 @@ defmodule Constable.Api.SearchController do
...
   alias Constable.Announcement
 
-  def create(conn, %{"query" => search_terms}) do
-    announcements = Announcement.search(search_terms) |> Repo.all
+  def create(conn, params = %{"query" => search_terms}) do
+    excludes = params["exclude_interests"] || []
+    announcements = 
+      search_terms
+      |> Announcement.search
+      |> Announcement.without_interests(excludes)
```
paulcsmith: `without_interests` seems like it could be stripping interests from the announcement. I think the wording in the param might make sense as the function name instead. Something like `Announcement.excluding_interests`? Not a big deal either way though. Just a thought

web/controllers/search_controller.ex
```diff
@@ -18,8 +18,11 @@ defmodule Constable.SearchController do
     render conn, "new.html", announcements: []
   end
 
-  defp matching_announcements(%{"query" => search_terms}) do
+  defp matching_announcements(params = %{"query" => search_terms}) do
+    excluded_interests = params["exclude_interests"] || []
+
     Announcement.search(search_terms)
+    |> Announcement.without_interests(excluded_interests)
```
paulcsmith: This should probably be moved to `Announcement.search` since it's always used. So maybe a second arg should be added so it's `Announcement.search(search_terms, exclude_interests: excluded_interests)`

web/models/announcement.ex
```diff
@@ -44,6 +44,12 @@ defmodule Constable.Announcement do
     ]
   end
 
+  def without_interests(query, interests \\ []) do
+    from a in query,
+      full_join: i in assoc(a, :interests),
+      where: not i.name in ^interests or is_nil(i.name)
```
paulcsmith: I'm a bit confused by the `is_nil(i.name)`. Is that saying it should also exclude announcements that on't have an interest at all?

test/controllers/api/search_controller_test.exs
BlakeWilliams: Ah yeah, I should have done that. 👍

web/models/announcement.ex
BlakeWilliams: Yeah, the query isn't super straightforward, it's basically saying "find all interests that don't have an excluded interest OR it has no interests", because it was filtering announcements that had 0 interests.

paulcsmith: Do we want to filter announcements with no interests though? It seems like that might be something we want still since sometimes people forget to add them. Other than that I think the query looks pretty darn good!

BlakeWilliams: That's a really good question, and I'm not sure I have a solid answer. I was thinking we could try including them for now and exclude them later if it ends up being a problem. Also thank you! :)

croaky:
Thanks, @BlakeWilliams!

Issue #325, opened 2016-06-07
Dates on posts


mjankowski:
The "x days ago" is fine ... but it would be nice to have the actual date visible somewhere.

When I'm digging back to see when something really old was posted, it's weird to do the mental "ok, what day was 89 days ago?" math...

PR #445, opened 2017-09-01
Add title to announcement published date

germsvel:
Taking a stab at https://github.com/thoughtbot/constable/issues/325. 

Summary
=======

Using the `time_ago_in_words/1` to display the published date is really nice since it shows something like, `"2 days ago"`. When the date is too far back, however, it can get confusing.

This commit adds the title attribute to the announcement date so that the announcement published date shows up with the format: `"yyyy-mm-dd"`. It is pretty rough, but it should do the job. 

Screenshot
========

screen shot 2017-09-01 at 4 28 15 pm


dbernheisel:
LGTM!

PR #447, opened 2017-09-15
Easily reactivate users

paulcsmith:
Users are deactivated if they leave thoughtbot. Sometimes someone with the same
first name joins thoughtbot later and inherits the deactivated email so we need
to reactivate the email.

This adds instructions and method for easily reactivating a user


paulcsmith:
@germsvel Yeah I agree that having a Users page where you can activate/deactivate would be better if this happens often!

PR #449, opened 2017-12-09
Fix bourbon deprecation warnings

germsvel:
When running mix tasks like `mix tests` or `mix phoenix.server`, bourbon throws a lot of deprecation warnings, like the following: 

screen shot 2017-12-09 at 1 55 48 pm

This commit fixes bourbon warnings for `display()`, `align-items()`, `animation()`, and `em()`.

For `em()`, we actually copy the core of the function and base value into a new function that we call `px-to-em()` so that we can continue using it in the single file where it is used. It may be worth in the future to use `em`s directly.

### DIFF
--- a/web/static/css/_comments.scss
+++ b/web/static/css/_comments.scss
@@ -10,12 +10,12 @@
 }
 
 .comment-loading {
-  @include animation(loading-spinner 0.6s infinite linear);
   @include border-style(solid);
   @include border-width(0.2em);
   @include hide-text();
   @include margin(null auto);
   @include size(2em);
+  animation: $long-duration linear infinite loading-spinner;
   border-color: $base-border-color;
   border-left-color: $dark-blue;
   border-radius: 50%;
--- a/web/static/css/base/_grid-settings.scss
+++ b/web/static/css/base/_grid-settings.scss
@@ -1,6 +1,6 @@
-$small-screen: new-breakpoint(max-width em(540));
-$medium-screen: new-breakpoint(max-width em(720));
-$large-screen: new-breakpoint(max-width em(860));
+$small-screen: new-breakpoint(max-width 540px);
+$medium-screen: new-breakpoint(max-width 720px);
+$large-screen: new-breakpoint(max-width 860px);
 
-$small-screen-up: new-breakpoint(min-width em(520));
-$medium-screen-up: new-breakpoint(min-width em(720));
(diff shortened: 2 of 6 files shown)


web/static/css/base/_grid-settings.scss
```diff
@@ -1,6 +1,12 @@
...
-$large-screen: new-breakpoint(max-width em(860));
+@function px-to-em($value) {
+  $em-base: 16;
 
-$small-screen-up: new-breakpoint(min-width em(520));
-$medium-screen-up: new-breakpoint(min-width em(720));
+  @return ($value / $em-base) * 1em;
+}
+
+$small-screen: new-breakpoint(max-width px-to-em(540));
```
mborsare: I'm not sure if there's a benefit to this function. Why not just use pixels for the breakpoints? That would solve the deprecation issue and not require any additional functions.

mborsare: i.e. `$small-screen: new-breakpoint(max-width 540px);`

web/static/css/base/_grid-settings.scss
```diff
@@ -1,6 +1,12 @@
-$small-screen: new-breakpoint(max-width em(540));
```
mborsare: `$small-screen: 540px`

web/static/css/base/_grid-settings.scss
germsvel: 👍 yeah that sounds much better. I'll update this.

web/static/css/_comments.scss
```diff
@@ -10,7 +10,7 @@
 }
 
 .comment-loading {
-  @include animation(loading-spinner 0.6s infinite linear);
+  animation: 0.6s linear infinite loading-spinner;
```
mborsare: Since 0.6s is used twice we should define it as a variable. How about something like `$long-duration: 0.6s` in `_variables.scss`?

mborsare: then `animation: $long-duration linear infinite loading-spinner;`

mborsare:
LGTM after that variable comment is addressed. 👍

web/static/css/_comments.scss
germsvel: Good idea. Extracted in 75bef98668c7d6429e4e593883c6640672913703

germsvel:
Thanks for the help @mborsare!

Issue #450, opened 2017-12-22
Upgrade to Phoenix 1.3.0


germsvel:
We should upgrade constable to use Phoenix 1.3.0. 

A guide for the upgrade can be found [here](http://phoenixframework.org/blog/upgrading-from-120-to-130) and a slightly older version [here](https://gist.github.com/chrismccord/71ab10d433c98b714b75c886eff17357). 

Phoenix v1.3.0 is supposed to be backwards compatible release with v1.2.x., so we could take this in steps. I think following the upgrade guide we could do it in these steps,

- [x] [bump dependency](https://gist.github.com/chrismccord/71ab10d433c98b714b75c886eff17357#bump-your-phoenix-dep) and [update static manifest location](https://gist.github.com/chrismccord/71ab10d433c98b714b75c886eff17357#update-your-static-manifest-location)
- [x] [update project to use new directory structure](https://gist.github.com/chrismccord/71ab10d433c98b714b75c886eff17357#update-your-project-the-new-13-directory-structure-optional)
- [x] [move static assets into `assets` directory](https://gist.github.com/chrismccord/71ab10d433c98b714b75c886eff17357#move-static-assets-inside-self-contained-assets-directory)
- [ ] In step 2, everything in the `web` directory would be moved into `lib/constable_web` which would include models. In this step, we should bring anything that is not a controller, view, or channel out of `lib/constable_web` and into `lib/constable`.

Of the three, I think updating the project to use the new structure will be the biggest change, and even though it's technically optional, I think we should move the project to have the format that phoenix projects will have in the future.

germsvel:
I hadn't closed this issue yet because the last step is not actually finished. That step is technically  not necessary for Phoenix 1.3.0, but I think it makes sense to finish updating the application to be a "standard" Phoenix 1.3 app.

mjankowski:
I opened https://github.com/thoughtbot/constable/pull/526 which I think might complete the last part here...

After those changes, the things left in `constable_web` are:

- The folders for: channels, controllers, templates, views
- controller_helper.ex (should this move?)
- emails.ex (move?)
- endpoint, gettext, router - seems these all stay here in 1.3 projects

mjankowski:
This MIGHT be done now after latest batch of merges.

germsvel:
@mjankowski sorry for taking so long to respond. I do think that with your latest batch of changes, we can call this done. 

[...]

PR #451, opened 2017-12-22
Bump Phoenix version to 1.3.0

germsvel:
Summary
=======

This is the first step in upgrading to Phoenix 1.3.0. We are following the instructions to upgrade phoenix found [here](https://gist.github.com/chrismccord/71ab10d433c98b714b75c886eff17357).

The first step outlined in https://github.com/thoughtbot/constable/issues/450 is to bump the phoenix version and to update the static manifest location. We do that here.

The bump in Phoenix version caused a warning for Phoenix.Token to show up. It now requires a `max_age` to be set when verifying a token. We set the max age to a week here.


drapergeek:
👍

PR #452, opened 2017-12-22
Re-introduce the log out button

tjmw:
Accessible from the settings modal. The route/controller action etc all already existed (#303) but the link was commented as part of the modal redesign (dc56f84a92). I'm not too sure if that was deliberate, perhaps I'm the only one who ever used it :-)

I tweaked the markup of the modal slightly to accommodate the new button.

Before:

screen shot 2017-12-22 at 16 01 34

After:

screen shot 2017-12-22 at 16 01 11

If anyone has thoughts on the styling let me know and I'll revisit.

### DIFF
--- a/web/templates/settings/show.html.eex
+++ b/web/templates/settings/show.html.eex
@@ -8,8 +8,8 @@
       </a>
     </header>
 
-    <%= form_for @changeset, settings_path(@conn, :update), [class: "settings-form"], fn(f) -> %>
-      <div class="modal-body">
+    <div class="modal-body">
+      <%= form_for @changeset, settings_path(@conn, :update), [class: "settings-form"], fn(f) -> %>
         <div class='field'>
           <label>Your name:</label>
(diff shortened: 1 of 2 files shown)


web/templates/settings/show.html.eex
```diff
@@ -39,13 +39,13 @@
...
-      <footer>
-        <%# <%= link "Logout of Constable", to: session_path(@conn, :delete), method: :delete %1> %>
         <%= submit gettext("Save") %>
-      </footer>
-    <% end %>
+      <% end %>
+    </div>
+
+    <footer>
+      <%= link "Logout of Constable", to: session_path(@conn, :delete), method: :delete, class: "button button-secondary" %>
```
germsvel: I know the text was already there, but I think we may want `Log out` since it's a verb as opposed to when its used as a noun (e.g. "Logout page").

germsvel:
Yeah, this looks good to me!

PR #453, opened 2017-12-22
Move /web to /lib/constable_web for Phoenix 1.3.0.

germsvel:
These are the second and third step outlined in https://github.com/thoughtbot/constable/issues/450 to update Constable to Phoenix 1.3.0. It follows very closely the upgrade instructions [found here](https://gist.github.com/chrismccord/71ab10d433c98b714b75c886eff17357)

Summary
=======

Phoenix 1.3.x applications have moved to a directory structure that more closely resembles other mix projects.

This means that we are moving top level `/web` directory to `/lib/constable_web`, and all controllers, channels, views, the router, and the endpoint are now namespaced by `ConstableWeb` instead of `Constable`. Essentially, anything that is related solely with the web application gets namespaced under `ConstableWeb`. Note how `Repo` does not get namespaced by `ConstableWeb` since it is not part of the web delivery mechanism.

Note that we are still leaving all models under the `/constable_web` directory. The fourth step of this process outlined in #450 will handle the moving of models and possible renaming of the directory.

We also move all assets to their own `assets` directory. This required moving all files previously under `web/static/assets` to `assets`, moving `package.json` and `brunch-config.js` into `assets`, updating all templates that use javascript to the point to the correct directory, and updating our `mix test` alias. 

Note that we included a new script `bin/test_suite` because our `mix test` task can no longer dynamically build the assets since brunch-config has been moved into assets. Brunch expects us to build brunch from the directory in which `brunch-config.js` is defined. This is not as easy when running tests, so we just use the `bin/test_suite` for that.


germsvel:
Since all the assets are already being moved in this PR, I wonder if I should just move them out to a top-level `assets` directory here. Maybe there's no sense in moving all of those files and then moving them back.

drapergeek:
>Since all the assets are already being moved in this PR, I wonder if I should just move them out to a top-level assets directory here. Maybe there's no sense in moving all of those files and then moving them back.


I agree with your assessment

germsvel:
@drapergeek I finally got all the javascript to work correctly when moving them to the top-level `assets` directory! This should complete parts 2 and 3 of the related issue.

drapergeek:
LGTM :shipit:

PR #454, opened 2018-03-19
Redesign subscribe and unsubscribe UI

mborsare:
This PR creates a more cohesive and easier to understand subscription component for users. Previously we had a couple of different styles around this action. One was under "Manage your interests" and was confusingly similar in visual style to the new announcement action. The other was on the announcement page and used hard to understand iconography. It wasn't immediately clear whether or not someone was subscribed the announcement.

Tangentially this PR unifies the display of `#` before interests in a couple of different places. It also cleans up some unnecessary styles by adding a utility class for a container with padding top.

Before:
screen shot 2018-03-19 at 5 04 15 pm

screen shot 2018-03-19 at 5 04 50 pm

After:
screen shot 2018-03-19 at 5 11 17 pm

screen shot 2018-03-19 at 5 21 13 pm

Before (no hashtag ☹️):
screen shot 2018-03-19 at 5 22 30 pm

After (hashtag 😄) :
screen shot 2018-03-19 at 5 22 39 pm

### DIFF
--- a/lib/constable_web/templates/announcement/show.html.eex
+++ b/lib/constable_web/templates/announcement/show.html.eex
@@ -10,14 +10,14 @@
       <%= if @subscription do %>
         <%= link to: announcement_subscription_path(@conn, :delete, @announcement.id),
           method: :delete,
-          class: "button button-unsubscribe",
+          class: "unsubscribe-to",
           data: [turbolinks: "refresh"] do %>
-          <%= gettext("Unsubscribe") %>
+          <%= gettext("Subscribed") %>
         <% end %>
       <% else %>
         <%= link to: announcement_subscription_path(@conn, :create, @announcement.id),
           method: :post,
-          class: "button button-subscribe",
+          class: "subscribe-to",
           data: [turbolinks: "refresh"] do %>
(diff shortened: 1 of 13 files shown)


lib/constable_web/templates/announcement/show.html.eex
```diff
@@ -10,15 +10,17 @@
       <%= if @subscription do %>
         <%= link to: announcement_subscription_path(@conn, :delete, @announcement.id),
           method: :delete,
-          class: "button button-unsubscribe",
+          class: "unsubscribe-to",
           data: [turbolinks: "refresh"] do %>
-          <%= gettext("Unsubscribe") %>
+          <img class="subscribe-icons" src="/images/icon-check-white.svg" />
```
enatario: May be good to set an empty `alt` for this and the other image so it's not read by screen readers. Alternatively, since it's just decoration, could be a `:before` element with empty content and bg image that calls that svg.

mborsare: Great points. I like the pseudo element idea. Let me give it a go.

PR #456, opened 2018-03-21
Ui refactor for the announcement page

mborsare:
This PR makes a few improvements to the previous announcement page design.

**Goal 1: Create a clear entry point**
The previous design lacked a clear entry point into the content. This PR solves that problem by using a bolder weight and larger size for the announcement's main heading.

**Goal 2: Achieve better typographic hierarchy**
The previous type hierarchy was too subtle. This PR solves that by creating more distinction between the announcement body copy and comment thread. It also tightens up the line-height and shortens the line-length on the announcement body. This creates more readable announcements for Constable users. We've also improved the spacing throughout the app to feel more solid.

**Goal 3: Clarify what subscribing to an announcement means**
The old design placed the subscribe button adjacent to the announcement's tags. This creates confusion around what the reader is subscribing to. Is is the message or the tags? Is it all of the tags or just one? To fix this we've moved the subscribe button below the title and adjacent to the author's name. We've also clarified the button's label to be more specific. It now says "Subscribe to thread."

**Goal 4: Increase Constable's friendliness**
The old design had avatars that were very small. Because of that each one looked more-or-less the same. This PR increases both an announcement author's avatar and a commenter's. This helps Constable readers recognize faces better.

image

### DIFF
--- a/assets/css/_comments.scss
+++ b/assets/css/_comments.scss
@@ -4,9 +4,8 @@
 }
 
 .edit-comment {
-  font-size: 13px;
-  color: 999;
-  margin-left: 5px;
+  font-size: $tiny-font-size;
+  margin-left: $tiniest-spacing;
 }
 
 .comment-loading {
@@ -23,10 +22,11 @@
 }
 
 .comments-list {
-  border-top: 5px solid $light-gray;
+  border-top: $base-border;
   list-style-type: none;
-  margin-top: $base-spacing;
+  margin-top: $small-spacing;
   padding-top: $base-spacing;
+  position: relative;
 }
 
 .comment-body {
@@ -50,20 +50,30 @@
 .avatar-rounded {
   @include size($rounded-avatar-size);
   border-radius: 50%;
-  margin-right: $tiny-spacing / 2;
+  vertical-align: middle;
+}
+
+.avatar-rounded-large {
+  @include size($rounded-avatar-size-large);
 }
 
 .author-information {
   @extend %align-center-flex;
+  @include padding($tiniest-spacing null $tiny-spacing);
+  justify-content: space-between;
+}
+
+.author {
+  flex: 1;
 }
 
 .comment-information {
-  @extend %align-center-flex;
   justify-content: space-between;
   margin-bottom: $tiny-spacing / 2;
 }
 
-.comment-new, .comment-edit {
+.comment-new,
+.comment-edit {
   border-top: $base-border;
   margin-bottom: $base-spacing;
   padding-top: $base-spacing;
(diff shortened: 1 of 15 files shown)


assets/css/_comments.scss
```diff
@@ -50,20 +60,28 @@
...
   @include size($rounded-avatar-size);
   border-radius: 50%;
-  margin-right: $tiny-spacing / 2;
+  margin-right: $tiny-spacing;
+  vertical-align: middle;
+}
+
+.avatar-rounded-large {
+  @include size($rounded-avatar-size-large);
+  margin-right: $small-spacing;
```
creuter: Should this be handled by the new media object instead?

assets/css/_comments.scss
```diff
@@ -50,20 +60,28 @@
...
 
 .author-information {
   @extend %align-center-flex;
 }
 
 .comment-information {
-  @extend %align-center-flex;
   justify-content: space-between;
   margin-bottom: $tiny-spacing / 2;
+
```
creuter: Extra whitespace

assets/css/_comments.scss
```diff
@@ -50,20 +60,28 @@
...
+  @extend %align-center-flex;
```
creuter: Is this a thing? I don't see it elsewhere in the code base. If it is a thing, should it be the first declaration?

creuter:
I dig 🚢 🚢 🚢 

Especially liking the improved contrast in text

assets/css/_comments.scss
mborsare: Not sure tbh. I think it's highlighted b/c CSS Comb grabbed it. I'll investigate.

assets/css/_comments.scss
```diff
@@ -4,8 +4,8 @@
 }
 
 .edit-comment {
-  font-size: 13px;
   color: 999;
```
creuter: 🤷‍♂️

assets/css/_comments.scss
```diff
@@ -23,10 +23,20 @@
...
 .comments-list {
-  border-top: 5px solid $light-gray;
+  border-top: $base-border;
   list-style-type: none;
   margin-top: $base-spacing;
   padding-top: $base-spacing;
+  position: relative;
+
+  &::before {
+    content: "Comments";
```
creuter: Just curious, why do you prefer this to an `<h2>` (or whatever) in the DOM? I imagine that would be more accessible.

mborsare: You're right. I more or less mocked this up in browser and left it. I'll transition to more accessible content.

assets/css/_comments.scss
mborsare: 🤔

mborsare: The best color.

assets/css/_comments.scss
mborsare: Nice call!

PR #457, opened 2018-04-03
MVP UI/UX refactor for subscribed interests bar

mborsare:
This PR moves the edit your interests feature exclusively into
the `Your interests` tab. This makes the experience viewing
`All announcements` cleaner and more focused. It does the same
for `Your interests` because the action of editing is now tied only to
that tab.

interests


mborsare:
@paulcsmith and @germsvel I'm accomplishing this with pure CSS and am assuming that's not the best way to do it. If you're ok with the design change would you mind helping to improve the code so it isn't purely presentational?

germsvel:
@mborsare I'd be happy to help with this. 

It seems like both of those assignment lists share the same template, so purely css might be okay? We could do some conditional check and render things based on which of those two announcements pages we're in, but not sure how to distinguish between them off the top of my head. Anyway, I'd be happy to help.

assets/css/_announcements-interests-nav.scss
```diff
@@ -1,6 +1,10 @@
 .segmented-control {
+  background-color: $white;
   border-bottom: $base-border;
   font-size: $small-font-size;
+  line-height: $base-spacing;
+  position: relative;
+  z-index: $base-z-index + 1;
```
enatario: Does it make sense to use a var for this instead? I just took a look at the z-indexes we have and seems like we need an overhaul/more variation, but that may be a PR for another time.

assets/css/_announcements-interests-nav.scss
```diff
@@ -36,14 +40,17 @@
...
 }
 
-.mine.selected {
-  + .select-interests-link,
-  .caret {
-    border-bottom: $dark-blue 1px solid;
+.interests-message {
+  border: $base-border;
+  text-align: center;
+  transform: translateY(-1px);
```
enatario: Reason for this? Weird alignment things? Just curious.

lib/constable_web/templates/announcement/index.html.eex
```diff
@@ -12,22 +12,36 @@
...
-  <p class="interests-message container">
-    <%= dngettext("interest-count", "You're subscribed to 1 interest.", "You're subscribed to %{count} interests.", interest_count_for(@current_user)) %>
-    <%= link to: interest_path(@conn, :index), data: [role: "view-all-interests"] do %>
-      <%= gettext "Manage your interests." %>
-    <% end %>
-  </p>
+  <%= if !@show_all do %>
+    <p class="interests-message <%= class_for("mine", @conn) %>">
+      <%= link to: interest_path(@conn, :index), data: [role: "view-all-interests"] do %>
+        <span class="subscribed-to">
```
enatario: Maybe unnest this and put that class on the `a` and remove the `span`? Not sure if that makes the styling janky.

enatario: Oh wait I see why you did this now. The edit text is a separate thing. Disregard my earlier comment :)

assets/css/_announcements-interests-nav.scss
mborsare: Your comment made me realize I don't need this. I was using it to cancel out the border top on the container. I should really just have a border bottom and kill this declaration instead.

assets/css/_announcements-interests-nav.scss
mborsare: Yup, I'll investigate and maybe refactor now.

enatario: This may be a little heavy-handed for this project, but here's a map I often use for z-indexes (mostly to keep things agnostic and in context). Values Eric Bailey originally concocted actually!

```sass
$z-index: (
  basement: -999999,
  layout: 1000,
  object: 2000,
  component: 3000,
  subcomponent: 3010,
  modal: 4000,
  submodal: 4010,
  ceiling: 999999,
);
```

And then use this function:
```sass
@function get-z-index($index) {
  @return map-get($z-index, $index);
}
```

In use:
```sass
.foo-modal {
  z-index: get-z-index("modal");
}
```

assets/css/_announcements-interests-nav.scss
mborsare: @enatario https://github.com/thoughtbot/constable/pull/457/commits/3220c7c8a65175f974937ba2e8e18c6f30e8ed4e

assets/css/base/_variables.scss
```diff
@@ -66,10 +65,8 @@ $long-duration: 0.6s;
 
 // Z-indeces
 $base-z-index: 0;
-$z-index-subscribed-tooltip: 2;
-$z-index-interests-overlay: 1;
-$z-index-interests-select: 2;
-$z-index-modal: 1;
+$mezzanine-z-index: 500;
```
enatario: love this name

assets/css/_announcements-interests-nav.scss
mborsare: @enatario thoughts on this approach
https://github.com/thoughtbot/constable/pull/457/commits/3220c7c8a65175f974937ba2e8e18c6f30e8ed4e

enatario: fab 🎉

mborsare: Oh I dig that map, though I agree it's overkill for this. I'm going to make a gist of that!

Issue #455, opened 2018-03-20
CircleCI fails on erlang dependency


mborsare:
![screen shot 2018-03-20 at 2 34 24 pm](https://user-images.githubusercontent.com/5566826/37675510-d07b1146-2c4b-11e8-9d14-9cccc6ce4b16.png)

PR #463, opened 2018-04-13
Use CircleCi 2

germsvel:
What changed?
=============

We've been having issues running our test suite in CircleCi. The version of asdf we were using was pretty old. Instead of investing time fixing that, it seemed like a good idea to just go ahead and switch to CircleCi 2. So this should fix https://github.com/thoughtbot/constable/issues/455. 

This removes the old circle config file and adds a new one for circle ci 2.


.circleci/config.yml
```diff
@@ -0,0 +1,79 @@
...
+          paths: "assets/node_modules"
+
+      - run:
+          name: Wait for DB
+          command: dockerize -wait tcp://localhost:5432 -timeout 1m
+
+      - run: bin/test_suite
+
+      - store_test_results:
+          path: _build/test/junit
```
dbernheisel: what's this doing?

.circleci/config.yml
```diff
@@ -0,0 +1,79 @@
...
+      - save_cache:
+          key: v1-npm-cache-{{ .Branch }}
+          paths: "assets/node_modules"
+      - save_cache:
+          key: v1-npm-cache
+          paths: "assets/node_modules"
+
+      - run:
+          name: Wait for DB
+          command: dockerize -wait tcp://localhost:5432 -timeout 1m
```
dbernheisel: 👍 I haven't needed to do this with Joydrive, but this is great.

.circleci/config.yml
```diff
@@ -0,0 +1,79 @@
...
+          paths: "_build"
+
+      - restore_cache:
+          keys:
+            - v1-npm-cache-{{ .Branch }}-{{ checksum "assets/package-lock.json" }}
+            - v1-npm-cache-{{ .Branch }}
+            - v1-npm-cache
+      - run:
+          working_directory: assets
+          command: npm install && node_modules/.bin/brunch build
```
dbernheisel: I don't know how important it is to enforce the `node` or `npm` versions across environments, but here you're using whatever circleci chose to install in their image, which doesn't look configurable, and _may_ not be a version of `npm` that looks at the `package-lock.json`. I figure if it's any version of `npm` that uses the `package-lock.json`, then it's safe; but if they have an older version that doesn't look at `package-lock.json`, then you might experience inconsistent browser-based tests because it's installing different versions that what's locked.

I took the extra steps in another project to install `asdf` during this setup process and only use it for `nodejs`, but it stinks to have to do that.

dbernheisel:
LGTM

.circleci/config.yml
germsvel: I grabbed that from the [elixir language guide](https://circleci.com/docs/2.0/language-elixir/) from circle ci. I believe that's where we'd store test results for Test Summary if we were using more than one container. We probably don't need it for now since we're only using one.

.circleci/config.yml
germsvel: Yeah, I was really trying to avoid having to install. I was assuming/hoping it would just use `package-lock.json` but you're right in saying that if the version is old and doesn't look at `pacakge-lock.json`, then this might cause problems. I'm not sure. I feel like I'd rather cross that bridge when we get there, but I'm open to changing this if you feel strongly against that approach.

dbernheisel: I agree with you. Deal with it if it's a problem, not before. :+1:

PR #465, opened 2018-04-13
Add slugs to announcements

germsvel:
What?
=====

We add slugs to announcements for better urls. We do so in a way that maintains backwards compatibility and allows for urls with ids and urls with the new slugs.

A note on terminology
=====================

We technically add two things that could be called "slugs". One is a column in the announcements table. The other is what is now at the end of an announcement path (e.g. "announcements/:slug").

The slugs in the database are a slugified version of the announcement's title. The slugs in the path are really a composite slug of the announcement id and the slug in the database field. So the slugs in the url have the form of `id-slugified-title'.

Details
========

We add slugs to announcements paths that have an `id-sluggified-title` format. We do this by implementing the Phoenix.Param protocol for announcement so that all path helpers automatically generate a slug that is a combination of the id and the slug.

We also add a columng "slug" to announcements table. We are not including an index because we do not intend to retrieve records by the slug. We'll retrieve them by id.

In order to persistent and associations easily with the new slugs, we use a custom ecto type that knows how to cast `id-sluggified-title` formats as well as simple integers. We set that type as the primary key in announcement so that ecto knows how to handle it.

Why this approach?
==================

The benefit of this approach is that we don't have to worry about what form the "id" will come in. It could be an integer or it could be the new slugs. It'll cast it when saving the id to the database and essentially throw away the slug (for the purposes of the primary id). This means that all old urls are automatically okay, while new urls will be supported. It also means that changing the title of the announcement will break any old links someone saved outside of constable.

### DIFF
--- a/lib/constable/permalink_type.ex
+++ b/lib/constable/permalink_type.ex
@@ -1,0 +1,28 @@
+defmodule Constable.PermalinkType do
+  @behaviour Ecto.Type
+
+  def type, do: :id
+
+  def cast(integer) when is_integer(integer) do
+    {:ok, integer}
+  end
+
+  def cast(string) when is_binary(string) do
+    case Integer.parse(string) do
+      {int, _} when int > 0 -> {:ok, int}
+      _ -> :error
+    end
+  end
+
+  def cast(_) do
+    :error
+  end
+
+  def dump(integer) when is_integer(integer) do
+    {:ok, integer}
+  end
+
+  def load(integer) when is_integer(integer) do
+    {:ok, integer}
+  end
+end
--- a/lib/constable_web/models/announcement.ex
+++ b/lib/constable_web/models/announcement.ex
@@ -6,9 +6,11 @@ defmodule Constable.Announcement do
   alias Constable.Interest
   alias Constable.AnnouncementInterest
 
+  @primary_key {:id, Constable.PermalinkType, autogenerate: true}
   schema "announcements" do
     field :title
     field :body
+    field :slug
     field :last_discussed_at, :utc_datetime, autogenerate: {DateTime, :utc_now, []}
     timestamps()
 
@@ -23,12 +25,14 @@ defmodule Constable.Announcement do
     announcement
     |> cast(params, ~w(title body))
     |> validate_required([:title, :body])
+    |> generate_slug()
   end
 
   def create_changeset(announcement, params) do
     announcement
     |> cast(params, ~w(title body user_id))
     |> validate_required([:title, :body])
+    |> generate_slug()
   end
 
   def last_discussed_first(query \\ __MODULE__) do
@@ -65,6 +69,17 @@ defmodule Constable.Announcement do
     )
   end
 
+  defp generate_slug(changeset) do
+    case get_change(changeset, :title) do
+      nil -> changeset
+      title -> put_change(changeset, :slug, slugify(title))
+    end
+  end
+
+  defp slugify(title) do
+    Slugger.slugify_downcase(title)
+  end
+
   defp prepare_for_tsquery(search_term) do
     search_term
     |> String.split(" ", trim: true)
@@ -74,5 +89,11 @@ defmodule Constable.Announcement do
 
   defp newest_comments_first do
     from(c in Comment, order_by: [asc: c.inserted_at])
+  end
+end
+
+defimpl Phoenix.Param, for: Constable.Announcement do
+  def to_param(%{slug: slug, id: id}) do
+    "#{id}-#{slug}"
   end
 end
--- a/priv/repo/migrations/20180413164621_add_slug_to_announcements.exs
+++ b/priv/repo/migrations/20180413164621_add_slug_to_announcements.exs
@@ -1,0 +1,9 @@
+defmodule Constable.Repo.Migrations.AddSlugToAnnouncements do
+  use Ecto.Migration
+
+  def change do
+    alter table(:announcements) do
+      add :slug, :string
+    end
+  end
+end
--- a/test/lib/constable/permalink_type_test.exs
+++ b/test/lib/constable/permalink_type_test.exs
@@ -1,0 +1,55 @@
+defmodule Constable.PermalinkTypeTest do
+  use ExUnit.Case, async: true
+
+  alias Constable.PermalinkType
+
+  test "type/0 is :id" do
+    assert :id == PermalinkType.type()
+  end
+
+  describe ".cast/1" do
+    test "casts integers" do
+      value = 1
+
+      cast_value = PermalinkType.cast(value)
+
+      assert cast_value == {:ok, value}
+    end
+
+    test "casts slugs" do
+      value = "1-some-title"
+
+      cast_value = PermalinkType.cast(value)
+
+      assert cast_value == {:ok, 1}
+    end
+
+    test "returns error if cannot cast slug" do
+      value = "some-title"
+
+      cast_value = PermalinkType.cast(value)
+
+      assert cast_value == :error
+    end
+
+    test "returns error for all other cases" do
+      value = %{}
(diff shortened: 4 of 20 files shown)


test/lib/constable/permalink_type_test.exs
```diff
@@ -0,0 +1,55 @@
...
+
+  test "type/0 is :id" do
+    assert :id == PermalinkType.type()
+  end
+
+  describe ".cast/1" do
+    test "casts integers" do
+      value = 1
+
+      casted_value = PermalinkType.cast(value)
```
danbee: This could just be `cast_value`.

germsvel: Thanks! `cast_value` is much better. It made me laugh to think of `casted_value` once you pointed it out. Updated in 7ae90b94221df17c57b8918a8e0ffc19e8dd65e0

test/models/announcement_test.exs
```diff
@@ -1,34 +1,89 @@
...
+        announcement
+        |> Announcement.update_changeset(%{title: title})
+        |> Changeset.get_change(:slug)
 
-    assert List.first(announcements).id == newest.id
-    assert List.last(announcements).id == oldest.id
+      assert slug == "a-normal-title"
+    end
+
+    test "does not generate slug if not title is nil" do
```
drapergeek: I think this has an extra "not" in it:

"does not generate slug if title is nil"

lib/constable/permalink_type.ex
```diff
@@ -0,0 +1,28 @@
+defmodule Constable.PermalinkType do
+  @behaviour Ecto.Type
+
+  def type, do: :id
+
+  def cast(integer) when is_integer(integer) do
+    {:ok, integer}
+  end
+
+  def cast(string) when is_binary(string) do
```
drapergeek: I'm sorry if I'm slow, but I can't seem to understand why the extra type here is necessary. Where is the app casting the `id` as a string with the slug?

lib/constable_web/models/announcement.ex
```diff
@@ -76,3 +91,9 @@ defmodule Constable.Announcement do
     from(c in Comment, order_by: [asc: c.inserted_at])
   end
 end
+
+defimpl Phoenix.Param, for: Constable.Announcement do
+  def to_param(%{slug: slug, id: id}) do
```
drapergeek: :+1: So nice!

priv/repo/migrations/20180413164621_add_slug_to_announcements.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Constable.Repo.Migrations.AddSlugToAnnouncements do
+  use Ecto.Migration
+
+  def change do
+    alter table(:announcements) do
+      add :slug, :string
```
drapergeek: What do you think about writing a functino to backfill these in this migration and then changing this to `null: false?

lib/constable/permalink_type.ex
germsvel: Yeah, this is a little confusing. Maybe I should do a better job of documenting this somehow in the code itself somehow. I'm open to ideas. 

Say you have an announcement with title `Elixir is awesome` and the id is `23`. Because we have defined the `Phoenix.Param` protocol as `"#{id}-#{slug}"`, the announcement url will look like this, `announcements/23-elixir-is-awesome`. 

When that hits the controllers, we just consider `23-elixir-is-awesome` as the `:id` since the route is `announcements/:id` (even though the whole slug is not the announcement's database id). So when we retrieve or insert any announcements, we are passing `%Announcement{"id" => "23-elixir-is-awesome"}` to ecto. 

Since we have defined the custom type, ecto will cast that id and use this function we're looking at in this comment (which is why we're checking for binary). At that point, it'll split the `"23-elixir-is-awesome"`, and just use the `23` which is the actual database id for getting/updating the database records. 

I am really open to suggestions on how to best tackle this, and if this is not the right solution for slugs, I'm also happy to scrap this altogether. I don't want to introduce confusing code if it's not needed. This was just the best way I could think of that's low effort to add ids. But it is a little bit magical because it's the combination of `Phoenix.Param` and this ecto type that are working in conjunction.

priv/repo/migrations/20180413164621_add_slug_to_announcements.exs
germsvel: I do like that idea. I know that in Rails when I'm going to do something like that I drop straight into sql so that the migration does not depend on whether or not the code modules exist (if someone runs this later and things have changed). But here I might want to use the `Slugger.slugify_downcase/1` function slugify the titles. 

Do you think that would be a bad idea?

drapergeek: I think given how simple it is, I'd just copy the code over. If that code evolves over time, it wouldn't hurt the migrations but if the original module disappeared, it makes life harder.

lib/constable/permalink_type.ex
drapergeek: This seems totally reasonable after your explanation.

The only other thing you may consider is that you don't need to always look it up by `Repo.get` and you could use a module for doing your lookups. The only benefit I see there is that your associations don't need to know that it's a special type.

germsvel: Hmmm. Something like that could really work. In a previous iteration, I tried doing something like that. In that iteration I didn't have the `id-slugified-title` format. I just had the `slug` so it was a pain when I had to create/update the associated models because they needed access to the id. So I would have to get the announcement by slug, in order to get the id, just so I could add/update a comment. But now, I could add a plug that stripped away the slug in the controllers that needed it, and just have the id there. That also separates the responsibilities better (in my opinion) since the slugs in the `id-slugified-title` are simply a web concern. 

I'll consider that, and it might actually simply this a lot more.

priv/repo/migrations/20180413164621_add_slug_to_announcements.exs
germsvel: Okay. That sounds good. I'll move the code that's in the mix task in here then, and hope that no one runs into issues. I hope we don't get rid of the `Announcements` module in an application about announcements. 😁

test/models/announcement_test.exs
germsvel: Updated in 3fac16c8443a6fd0449f449569ad8b716a5d86cd

lib/constable/permalink_type.ex
germsvel: I gave this a shot. I'm actually not sure which I prefer. They both seem okay. If you're interested in this, I have separate branch with the alternate implementation which uses a `plug` to de-slugify the slug before hitting the controllers. 

The comparison with this branch is here -> https://github.com/thoughtbot/constable/compare/gv-url-id-slugs...gv-slug-ids-plug

germsvel:
Closing this in favor of https://github.com/thoughtbot/constable/pull/489.

lib/constable/permalink_type.ex
germsvel: @drapergeek I ended up going with a different approach from what I had here with the custom ecto type. I started feeling like this approach coupled a view concern (the url having a slug) with how we were persisting the data in the database. I didn't really like that, so I just went with a plug that removes the slugified part. If you're interested in looking at that, it's on #489. Everything else in that PR is the same as it was here.

priv/repo/migrations/20180413164621_add_slug_to_announcements.exs
germsvel: The work is in a different branch, since I'm going with a different approach, but I addressed this comment in that branch (See https://github.com/thoughtbot/constable/commit/5031f773d20ee23cc243a178ad3a7b6604a72950).

Issue #493, opened 2018-06-08
Delete Option?


ThoughtbotThom:
Would love the ability for a user to delete their post, either as an option next to the edit link, or within the editor view would work

germsvel:
@ThoughtbotThom I am going to close this issue since I think it was addressed by #497 (which made it to production recently). Feel free to reopen if it shouldn't have been closed!

PR #497, opened 2018-07-13
Users can delete their own announcements

setphen:
addresses #493 

* If a user owns an announcement, allow them to delete it.
* Improve the styling of inline actions

_Help wanted: I want to write some tests for this, but Wallaby is very out of date and I don't know where to find the old documentation. Perhaps this can wait to be merged until we update Wallaby._

---

**before**

screenshot_2018-07-13 constable - testing 1 2 3

**after** 

screenshot_2018-07-13 constable - a very good announcement

### DIFF
--- a/lib/constable_web/controllers/announcement_controller.ex
+++ b/lib/constable_web/controllers/announcement_controller.ex
@@ -109,6 +109,22 @@ defmodule ConstableWeb.AnnouncementController do
     end
   end
 
+  def delete(conn, %{"id" => id}) do
+    current_user = current_user(conn)
+    announcement = Repo.get!(Announcement, id)
+
+    if announcement.user_id == current_user.id do
+      Repo.delete!(announcement)
+      conn
+      |> put_flash(:info, gettext("Announcement deleted"))
+      |> redirect(to: announcement_path(conn, :index))
+    else
+      conn
+      |> put_flash(:error, gettext("You do not have permission to delete that announcement"))
+      |> redirect(to: announcement_path(conn, :show, announcement))
+    end
+  end
+
   defp render_form(conn, action, announcement) do
     changeset = Announcement.create_changeset(announcement, %{})
     interests = Repo.all(Interest)
--- a/lib/constable_web/models/announcement.ex
+++ b/lib/constable_web/models/announcement.ex
@@ -16,7 +16,7 @@ defmodule Constable.Announcement do
     belongs_to :user, User
     has_many :comments, Comment, on_delete: :delete_all
     has_many :subscriptions, Subscription, on_delete: :delete_all
-    many_to_many :interests, Interest, join_through: AnnouncementInterest
+    many_to_many :interests, Interest, on_delete: :delete_all, join_through: AnnouncementInterest
     has_many :interested_users, through: [:interests, :users]
   end
 
--- a/lib/constable_web/templates/announcement/show.html.eex
+++ b/lib/constable_web/templates/announcement/show.html.eex
@@ -9,8 +9,16 @@
   <h1 data-role="title">
     <%= @announcement.title %>
     <%= if @announcement.user_id == @current_user.id do %>
-      <small>
+      <small class="inline-action">
         <%= link "edit", to: announcement_path(@conn, :edit, @announcement), data: [role: "edit"] %>
+      </small>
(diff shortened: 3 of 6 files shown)


germsvel:
@setphen yeah, wallaby being so out of date makes it hard to update. If you want to add tests in the meantime, I tend to look at the README for the 0.7.0 version -> https://github.com/keathley/wallaby/tree/v0.7.0. Not ideal, but tends to work until we update it to 0.20.0.

lib/constable_web/controllers/announcement_controller.ex
```diff
@@ -109,6 +109,22 @@ defmodule ConstableWeb.AnnouncementController do
...
   end
 
+  def delete(conn, %{"id" => id}) do
+    current_user = current_user(conn)
+    announcement = Repo.get!(Announcement, id)
+
+    if announcement.user_id == current_user.id do
+      Repo.delete!(announcement)
+      conn
+      |> put_flash(:info, gettext("Announcement deleted"))
```
germsvel: 👍 nice use of `gettext` for I18n.

lib/constable_web/templates/announcement/show.html.eex
```diff
@@ -9,9 +9,17 @@
...
-      <small>
+      <small class="inline-action">
         <%= link "edit", to: announcement_path(@conn, :edit, @announcement), data: [role: "edit"] %>
       </small>
+      <small class="inline-action">
+        <%= link(
+            "delete",
+            to: announcement_path(@conn, :delete, @announcement),
+            method: :delete,
+            data: [confirm: "Really delete this announcement?", role: "delete"]
```
germsvel: Should we internationalize the confirm message?

lib/constable_web/controllers/announcement_controller.ex
```diff
@@ -109,6 +109,22 @@ defmodule ConstableWeb.AnnouncementController do
     end
   end
 
+  def delete(conn, %{"id" => id}) do
```
germsvel: What do you think about adding controller tests for this action? Not sure if it's needed or not, but it might be good to have something in lieu of not having feature tests that test this path.

lib/constable_web/models/announcement.ex
```diff
@@ -16,7 +16,7 @@ defmodule Constable.Announcement do
     belongs_to :user, User
     has_many :comments, Comment, on_delete: :delete_all
     has_many :subscriptions, Subscription, on_delete: :delete_all
-    many_to_many :interests, Interest, join_through: AnnouncementInterest
+    many_to_many :interests, Interest, on_delete: :delete_all, join_through: AnnouncementInterest
```
germsvel: Do you know if this deletes all the associated `announcements_interests` records associated or does it delete all of the associated `interests` as well? We probably want the former but definitely not the latter, right?

test/controllers/announcement_controller_test.exs
```diff
@@ -1,5 +1,6 @@
 defmodule ConstableWeb.AnnouncementControllerTest do
   use ConstableWeb.ConnCase, async: true
+  import ConstableWeb.ControllerHelper, only: [current_user: 1]
```
germsvel: Is this import needed? I may have missed it, but I didn't see `current_user/1` being used.

lib/constable_web/models/announcement.ex
germsvel: It seems I should've done some reading before. It seems like it won't delete interests, 

> If you attempt to remove associated many_to_many data, Ecto will always remove data from the join schema and never from the target associations be it by setting :on_replace to :delete, :on_delete to :delete_all or by using changeset functions such as Ecto.Changeset.put_assoc/3. For example, if a Post has a many to many relationship with Tag, setting :on_delete to :delete_all will only delete entries from the “posts_tags” table in case Post is deleted.

from https://hexdocs.pm/ecto/Ecto.Schema.html#many_to_many/3-removing-data

setphen: Correct, it doesn't delete the interests, just the association.

test/controllers/announcement_controller_test.exs
setphen: Ah, I was trying to use this, but found a better way to write the test.

PR #500, opened 2018-07-20
Improve pagination controls

setphen:
The arrows of both pagination buttons are facing the same way 😆 
Let's also use the same pagination controls for announcements on the `interest/show` page as well.

screen shot 2018-07-20 at 10 14 28 am

### DIFF
--- a/lib/constable_web/templates/search/new.html.eex
+++ b/lib/constable_web/templates/search/new.html.eex
@@ -11,7 +11,7 @@
 
     <div class="container pagination">
       <%= if !on_first_page?(@search_page) do %>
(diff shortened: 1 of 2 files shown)


lib/constable_web/templates/search/new.html.eex
```diff
@@ -11,7 +11,7 @@
 
     <div class="container pagination">
       <%= if !on_first_page?(@search_page) do %>
-        <%= link gettext("foo"), to: search_path(@conn, :show, query: @query, page: (@search_page.page_number - 1)), class: "page-next button" %>
```
germsvel: 👍  nice catch!

PR #503, opened 2018-08-17
Update Wallaby to 0.20.x

germsvel:
What changed?
============

Our version of Wallaby is fairly outdated. We update to the newest version `0.20.0`.

The biggest changes that we made include the `Wallaby.Query` module. This module exposes functions like `link/1`, `button/1`, etc. that are meant to play nicely with `Wallaby.Browser` functions, such as `click/2`.

So we can do things like,

```elixir
session
|> fill_in(text_field("Name"), with: "thoughtbot")
|> click(button("Submit"))
```

where `fill_in/3` and `click/2` come from the browser module and `text_field/1` and `button/1` come form the query module.

Other changes:

The newest version of Wallaby also re-raises any javascript errors. For the longest time, `Shubox` has been undefined in development and testing, but now that Wallaby re-raises that error we had to handle it. I simply check to see if it's defined since the constant `Shubox` comes in through a script included in a header (though we may want to do something else).

### DIFF
--- a/assets/js/textarea-image-uploader.js
+++ b/assets/js/textarea-image-uploader.js
@@ -12,5 +12,7 @@ export function setupImageUploader(selector) {
     },
   }
 
-  new Shubox(selector, shuboxOptions);
+  if (typeof(Shubox) !== 'undefined') {
+    new Shubox(selector, shuboxOptions);
+  }
 }
--- a/test/acceptance/user_announcement_test.exs
+++ b/test/acceptance/user_announcement_test.exs
@@ -1,14 +1,17 @@
 defmodule ConstableWeb.UserAnnouncementTest do
   use ConstableWeb.AcceptanceCase
 
+  @announcement_title text_field("announcement_title")
+  @announcement_body text_field("announcement_body")
+
   test "user creates an announcement", %{session: session} do
     user = insert(:user)
 
     session
     |> visit(announcement_path(Endpoint, :new, as: user.id))
-    |> fill_in("announcement_title", with: "Hello World")
+    |> fill_in(@announcement_title, with: "Hello World")
     |> fill_in_interests("everyone")
-    |> fill_in("announcement_body", with: "# Hello!")
+    |> fill_in(@announcement_body, with: "# Hello!")
     |> click_submit_button
 
     assert has_announcement_title?(session, "Hello World")
@@ -25,7 +28,7 @@ defmodule ConstableWeb.UserAnnouncementTest do
     session
     |> visit(announcement_path(Endpoint, :new, as: current_user.id))
     |> fill_in_interests("elixir")
-    |> click_link("2 people are subscribed")
+    |> click(link("2 people are subscribed"))
 
     assert has_recipient_preview?(session, "Blake, Paul")
   end
@@ -37,9 +40,9 @@ defmodule ConstableWeb.UserAnnouncementTest do
     session
     |> visit(announcement_path(Endpoint, :show, announcement, as: user.id))
     |> click_edit
-    |> fill_in("announcement_title", with: "Updated")
+    |> fill_in(@announcement_title, with: "Updated")
     |> fill_in_interests("updated")
-    |> fill_in("announcement_body", with: "# Updated")
+    |> fill_in(@announcement_body, with: "# Updated")
     |> click_submit_button
 
     assert has_announcement_title?(session, "Updated")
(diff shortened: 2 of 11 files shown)


assets/js/textarea-image-uploader.js
```diff
@@ -12,5 +12,7 @@ export function setupImageUploader(selector) {
     },
   }
 
-  new Shubox(selector, shuboxOptions);
+  if ('undefined' !== typeof(Shubox)) {
```
germsvel: I'd be very happy to consider alternatives to this way of dealing with the issue that `Shubox` is undefined in the test and development environments. I'm not super happy with this.

test/acceptance/user_announcement_test.exs
```diff
@@ -6,9 +6,9 @@ defmodule ConstableWeb.UserAnnouncementTest do
 
     session
     |> visit(announcement_path(Endpoint, :new, as: user.id))
-    |> fill_in("announcement_title", with: "Hello World")
+    |> fill_in(announcement_title(), with: "Hello World")
```
danbee: I would prefer to see this reference the field by its label. I noticed though that the form does not have labels. We should add some.

test/acceptance/user_announcement_test.exs
```diff
@@ -54,53 +54,56 @@ defmodule ConstableWeb.UserAnnouncementTest do
...
-    |> fill_in("announcement_body", with: "# Updated")
+    |> fill_in(announcement_title(), with: "Updated title")
+    |> fill_in(announcement_body(), with: "# Updated")
     |> click_submit_button
 
     assert has_announcement_title?(session, "Updated title")
     assert has_announcement_interest?(session, "elixir")
   end
 
+  defp announcement_title  do
```
danbee: Can these be constants too, like in `test/acceptance/user_manages_interests_test.exs`?

test/acceptance/user_announcement_test.exs
danbee: This might be better off as another pull request though.

assets/js/textarea-image-uploader.js
danbee: Perhaps get a review from someone in #javascript?

test/acceptance/user_announcement_test.exs
germsvel: Yeah, I thought about making them all like module attributes, but I wasn't sure if it made more sense like that or just as functions. I'll update these.

test/acceptance/user_announcement_test.exs
germsvel: Yeah, I was noticing that finding elements by css seems to be the way all of the acceptance tests are written.

assets/js/textarea-image-uploader.js
germsvel: Good idea.

test/acceptance/user_announcement_test.exs
germsvel: Updated these in 847c0eb5d274a83e9a1075ec1b0af7a5c7d4753f

assets/js/textarea-image-uploader.js
joshuaclayton: In my experiences, I've always seen this flipped, e.g. `typeof(Shubox) !== "undefined"`

germsvel: @MattMSumner `Shubox` gets defined by a [script we inject in the layout template](https://github.com/thoughtbot/constable/blob/master/lib/constable_web/templates/layout/app.html.eex#L14). I think it is defined in development (I was incorrect on that account) but it is not defined in the test because we use a dummy url (in https://github.com/thoughtbot/constable/blob/master/config/test.exs#L15)

@joshuaclayton that's helpful. But you think use `typeof(Shubox)` is okay instead of creating a null object or something like that?

germsvel: Thanks for the feedback. I flipped the conditional in 4bee967445568af337167fdfa4d42d549d549af5

joshuaclayton: @germsvel my guess is that the page would still work without the JS addition; is that correct? If so, I'd avoid a shim/null object and allow the browser to handle it the old-fashioned way.

If there *is* some sensible fallback behavior (that's still handled by JS), a null object seems like a reasonable approach, but perhaps better suited for a separate PR anyway. Thoughts?

germsvel: Yeah, that makes sense. The page works totally fine without `Shubox` being there, so there's no harm in leaving it like it is. And you're right, if there is some fallback behavior, I can introduce it in a separate PR. 

Thanks for the feedback!

danbee: What is Shubox? Is it this? https://shubox.io

germsvel: Yeah, that's the one. A former thoughtbotter's side project.

PR #514, opened 2018-08-31
Update OAuth2 to 0.8

germsvel:
What changed?
==============

The api changed in from 0.6 to 0.7, with backward incompatible changes. Fortunately, only a couple of things have affected constable:

* There is an issue with `get_token!` where we have to specifically `put_param(:client_secret, secret)` right before we use `OAuth2.Client.get_token!`. I am unsure as to why it is not able to use the `client_secret` set in the `client/1` function that we define in `GoogleStrategy`.

* We update `Pact`'s "request_with_access_token" module from to `OAuth2.Client` (instead of `Token`) since the `get` function was moved.

For more changes in 0.7, see [the changelog](https://github.com/scrogson/oauth2/blob/master/CHANGELOG.md#v070-2016-08-16).

Other notes:
============

* There are more backwards incompatible changes in 0.9, but we leave that for another commit.

* This upgrade is important because other libraries require `Poison` to be `~> 3.0`. By getting OAuth to 0.8 we can finally use `Poison` 3.1.0.


test/controllers/announcement_controller_test.exs
```diff
@@ -79,7 +79,7 @@ defmodule ConstableWeb.AnnouncementControllerTest do
     assert interest_names == ["boston", "everyone"]
   end
 
-  test "#delete does not work for users who are not the owner of the announcement", %{conn: conn, user: user} do
+  test "#delete does not work for users who are not the owner of the announcement", %{conn: conn} do
```
stratigos: Was this change just to get rid of an unused param warning? Just curious :)

germsvel: Yeah, since I was upgrading versions, I was paying attention to all the warnings and this one was one of them. So I just fixed it here.

PR #531, opened 2018-09-11
Phoenix defaults update

mjankowski:
This is some WIP for things I think might finish up https://github.com/thoughtbot/constable/issues/450

I'm most worried about the changes related to hostnames/posts/http/etc type settings ... they were improved in terms of getting things from env, but I want to get the PR app running to see if things are ok.


config/config.exs
```diff
@@ -5,18 +5,17 @@
...
 
+# General application configuration
+config :constable,
+  ecto_repos: [Constable.Repo]
+
 # Configures the endpoint
 config :constable, ConstableWeb.Endpoint,
   url: [host: "localhost"],
   secret_key_base: "tJ+MdrPlKWMpmz7JyJgSu/11xvwnNZo7Sz8IAacy9MM6di3GqackE9iNjhkHI9p8",
-  debug_errors: false,
```
mjankowski: This is a default now

config/config.exs
```diff
@@ -5,18 +5,17 @@
...
-  root: Path.expand("..", __DIR__),
-  code_reloader: false,
```
mjankowski: This is a default now

lib/constable/application.ex
```diff
@@ -1,27 +1,28 @@
...
-    import Supervisor.Spec, warn: false
+    import Supervisor.Spec
 
     unless Mix.env == "production" do
       Envy.auto_load
     end
 
     setup_dependencies()
 
+    # Define workers and child supervisors to be supervised
```
mjankowski: I don't understand the full ramifications of changing from `worker` to `supervisor` in this section, and am somewhat blindly following the new app template/style. I assume this was an elixir or phoenix change which enables this to be ok...

config/config.exs
germsvel: Is this a `default: true` in `config/dev.exs` or something like that?

lib/constable/application.ex
germsvel: Yeah, a quick glance at what ecto brings in via `use Ecto.Repo` (which we use in `Constable.Repo`) shows that it's a supervisor. So I'm guessing that, it either changed from a single process to being a supervisor or using `worker` here before was incorrect. 

Looking at `Phoenix.Endpoint`, the same thing is true. So I think the change is ok.

lib/constable_web/views/error_view.ex
```diff
@@ -1,16 +1,16 @@
...
-    "Server internal error - 500"
-  end
+  # If you want to customize a particular status code
+  # for a certain format, you may uncomment below.
+  # def render("500.html", _assigns) do
+  #   "Internal Server Error"
+  # end
 
-  # Render all other templates as 500
-  def render(_, assigns) do
```
mjankowski: I think that a side of this change will be to handle this better - https://app.honeybadger.io/projects/48229/faults/39407168

Issue #421, opened 2016-10-06
Include the full text of announcements and comments in the digest email


mjankowski:
Currently the digest includes the titles of announcements and a count of the comments left ... but does not include the announcement body or comment body text.

My preference would be to have a "kitchen sink" approach in the digest ... send me literally everything that happened on constable in the last day in an email each morning.

BlakeWilliams:
I thought about this the other morning when I woke up and agreed with this. I think this would be nice as an "enhanced" version of the digest.

mjankowski:
I will give someone a personally autographed picture of me if they do this.

BlakeWilliams:
When I reflect back on my life, not completing this task and getting the autographed picture is one of my largest regrets.

mjankowski:
I made that promise almost two years ago and it seems like I'm on the verge of giving this signed picture to myself. I'll share a picture of me receiving the picture for everyone else to enjoy.

PR #532, opened 2018-09-12
Emails digest includes more content

mjankowski:
Resolves https://github.com/thoughtbot/constable/issues/421

### DIFF
--- a/lib/constable/factory.ex
+++ b/lib/constable/factory.ex
@@ -86,7 +86,7 @@ defmodule Constable.Factory do
 
   def comment_factory do
     %Constable.Comment{
-      body: "Post Body",
+      body: "Comment Body",
       user: build(:user),
       announcement: build(:announcement)
     }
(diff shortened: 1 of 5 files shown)


mjankowski:
Before:

screen shot 2018-09-13 at 9 39 27 am

After:

screen shot 2018-09-13 at 10 08 54 am

I could see making this look better in future improvements, but this seems fine to get the information in there in the first place.

paulcsmith:
I like this, but I also liked how before you could quickly glance at everything that was added and discussed on one screen, even on mobile. What do you think about leaving the current section up top, and adding the full content underneath. Kind of like a table of contents at top with the full content you can jump to if you're interested?

mjankowski:
Like a quick summary of interests and announcements up top - and then a fuller complete digest of all post body content and comments down below? I like that, and will take a crack at that here and update.

paulcsmith:
Yep, exactly!

mjankowski:
screen shot 2018-09-13 at 11 18 53 am

paulcsmith:
Very nice! One tiny suggestion: I think it would look better to put "Posted by <user>" above the announcement body. I think that lines up a bit better. LGTM either way though

mjankowski:
Thing I merged looks like this..

screen shot 2018-09-13 at 12 32 45 pm

...which seems good enough, and can be improved further if wanted.

PR #535, opened 2018-09-14
Use time-elements to replace timestamps with relative time

mjankowski:
Resolves https://github.com/thoughtbot/constable/issues/325

As far as I can tell, this is what GitHub is actually using on their pages.

It seems like the simple js-replacement of the timestamp with a relative time works correctly in modern browsers as I've done this here. However, the replacement of content which was in the custom element will only work with polyfills in place for some browsers. I opted here to just not include any content to replace, because I couldn't figure out how/where to insert the custom elements polyfill.

### DIFF
--- a/lib/constable_web/templates/announcement/_comment.html.eex
+++ b/lib/constable_web/templates/announcement/_comment.html.eex
@@ -7,11 +7,9 @@
   <div class="media-body">
     <div class="author-information">
       <h4 class="author"><%= @comment.user.name %></h4>
-      <time class="comment-time">
-        <a href="#comment-<%= @comment.id %>">
-          <%= time_ago_in_words @comment.inserted_at %>
-        </a>
-      </time>
+      <a href="#comment-<%= @comment.id %>" class="comment-time">
+        <%= relative_timestamp(@comment.inserted_at) %>
+      </a>
 
       <%= if @comment.user_id == @current_user.id do %>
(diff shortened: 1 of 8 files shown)


lib/constable_web/views/shared_view.ex
```diff
@@ -16,6 +16,10 @@ defmodule ConstableWeb.SharedView do
     gravatar_url(user.email, secure: true)
   end
 
+  def datetime_string(datetime) do
+    datetime |> NaiveDateTime.to_iso8601
```
germsvel: I think it would be preferable to write this as `NaiveDateTime.to_iso8601(datetime)` to avoid the pipe operator when there's only one value.

lib/constable_web/templates/announcement/_comment.html.eex
```diff
@@ -7,11 +7,10 @@
   <div class="media-body">
     <div class="author-information">
       <h4 class="author"><%= @comment.user.name %></h4>
-      <time class="comment-time">
-        <a href="#comment-<%= @comment.id %>">
-          <%= time_ago_in_words @comment.inserted_at %>
```
germsvel: Do you know if we can remove the `time_ago_in_words` function if we're no longer using it? It might just be dead code after this change.

lib/constable_web/views/shared_view.ex
mjankowski: will do

lib/constable_web/templates/announcement/_comment.html.eex
mjankowski: I think between this and the other one we might be done with it ... I'll prob merge that one (email one) first and then see if I can kill it here.

Issue #401, opened 2016-08-01
Remove comment border when no comments are present


ryancoughlin:
![screen shot 2016-08-01 at 2 18 36 pm](https://cloud.githubusercontent.com/assets/1182341/17304288/e3df672a-57f2-11e6-86d6-a58374673035.png)

PR #536, opened 2018-09-14
Dont generate comments list unless there are comments

mjankowski:
Resolves https://github.com/thoughtbot/constable/issues/401 more or less.

screen shot 2018-09-14 at 4 50 29 pm

Showing the `ul` conditionally was disruptive to JS on the page which is listening to and updating that element, so I changed the order of styling instead.

### DIFF
--- a/assets/css/_comments.scss
+++ b/assets/css/_comments.scss
@@ -22,7 +22,6 @@
 }
 
 .comments-list {
-  border-top: $base-border;
   list-style-type: none;
   margin-top: $small-spacing;
   padding-top: $base-spacing;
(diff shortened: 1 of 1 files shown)


assets/css/_comments.scss
```diff
@@ -35,11 +34,10 @@
 
 .comment {
   @include padding($base-spacing null);
-  border-bottom: $base-border;
+  border-top: $base-border;
```
germsvel: Can the `&last-child` `border-bottom: none;` below (line 44) be removed now that this has changed?

Issue #448, opened 2017-10-05
Demo


ACPK:
The demo requires a Thoughtbot email address. Would it be possible to get a demo that anyone could see, or at least screenshots of how it works?

mjankowski:
I reset the demo server and confirmed that gmail.com emails are allowed there now.

PR #544, opened 2018-09-20
Permitted domain email fixes

mjankowski:
Resolves https://github.com/thoughtbot/constable/issues/448

It appears that despite having an env var set with the permitted email domain on the demo server ... the code does not actually use that env var. We'll have to set this on staging and production as well prior to deploy.


mjankowski:
When you auth with wrong domain:

screen shot 2018-09-20 at 12 10 12 pm]

Logs capture:

screen shot 2018-09-20 at 12 10 33 pm

PR #548, opened 2018-09-21
Update `Oauth2.Client.post` usage to match implementation

mjankowski:
I'm not sure if this never worked or if the signature changed in a library
upgrade on the OAuth2 side. In our tests we replace the entirety of the
`google_strategy` with dummy modules, so this never actually gets hit.

Open to feedback on how to update tests so that this would have been executed and caught there...?


mjankowski:
I think we'd need to swap in whatever the equivalent of a `FakeWeb` (ruby lib) is so that the requests were actually being made, instead of swapping in the fake classes via Pact...? Not sure if something like fakeweb exists in elixir/phoenix world.

PR #561, opened 2018-10-27
Use webpack instead of brunch

mjankowski:
This is a WIP PR against the phoenix 1.4 branch, not ready for merge until that one is ready.


test/acceptance/user_searches_announcements_test.exs
```diff
@@ -1,5 +1,5 @@
 defmodule ConstableWeb.UserSearchesAnnouncementsTest do
-  use ConstableWeb.AcceptanceCase, async: true
```
germsvel: Were the specs failing? Is that why we're removing the `async: true` here?

mjankowski: Yeah, it fails while there and I don't understand why. All the docs I can find about this instruct me to do exactly what we're doing re: async working ... so I'm not sure what to try next, and just removed it. I'd love to understand what's happening and leave it.

PR #603, opened 2019-04-20
Add thoughtbot design system

tysongach:
I haven’t been able to log in locally, so have only tested the sign in page…

## Before

Screen Shot 2019-04-20 at 14 46 35

## After

Screen Shot 2019-04-20 at 14 46 19


mjankowski:
In order to sign in locally, make sure the `CLIENT_ID` and `CLIENT_SECRET` values are set in the local `.env` file. You can grab values from the heroku staging app. With that in place, running `mix phx.server` on port 4000 - you should be able to do google oauth starting from localhost:4000

tysongach:
@mjankowski That all looks good on my end, and I'm able to visit the sign in screen locally, then be taken to Google for auth…when it spits me back to the app, I'm back on the sign in screen with a `You must sign up with a email address ` error message.

mjankowski:
You're doing this is an excellent source of "we should write this down".

Did you copy `.sample.env` to `.env`? If not, do so. Once you do that (or already have) - check in `.env` to make sure `PERMITTED_EMAIL_DOMAIN` is set to "thoughtbot.com" (I think sample has it as default).

tysongach:
Yep, that is all set in my `.env`. FWIW, the `bin/setup` ran fine and did all this, I believe.

mjankowski:
That error message shows up when the domain of the user authing does not match the domain of the env var, and based on your error message I suspect the env is not loaded correctly - it would be like "You must sign up with a thoughtbot.com email address" if it were.

Next guess:

- Try running with `heroku local -p 4000` instead of just `mix phx.server` to see if the heroku/foreman combination loads the env?
- Double check that the env files have the correct domain?
- Stop and restart server to make sure it has those changes?

Beyond that I'm not sure but can help tomorrow.

tysongach:
This having the same issue. 😕 

I’m in a sprint this week. I’ll get some help next week when I have the headspace.

entcheva:
I'm having the same issue. 

`heroku local -p 4000` returns `[OKAY] Loaded ENV .env File as KEY=VALUE Format`, and running `System.get_env("PERMITTED_EMAIL_DOMAIN")` in the console returns `"thoughtbot.com"`.

I'm on client work today, but can help dig deeper on an investment day in the future.

germsvel:
Still unsure what's going on here, but @tysongach and I discovered a temporary fix for working locally. 

In `config/config.exs` replace `System.get_env("PERMITTED_EMAIL_DOMAIN")` with `"thougthbot.com"` as the third argument in `config :constable, :permitted_email_domain`,

```diff
- config :constable, :permitted_email_domain, System.get_env("PERMITTED_EMAIL_DOMAIN")
+ config :constable, :permitted_email_domain, "thoughtbot.com"
```

Issue #608, opened 2019-05-09
Images don't have alt tags


enatario:
Images should have `alt` tags associated with them.

* Avatars should have an alt tag that says something along the lines of `picture of [person's name]`.

* Images used in a post or in comments should have an alt tag. We use markdown to place images which comes with the ability to add an alt description (`![description of image](imgurl)`) but it does not retain the description as an alt when posted.

mjankowski:
Adding the `alt` tags for avatars is easy enough and I'll do that now.

The issue with stripping things from markdown is that the sanitizer we're using has various modes and I'm struggling to find one which allows both iframes and alt tags at the same time. Will keep looking.

enatario:
Awesome! Thank you for looking into it! 🤗

PR #609, opened 2019-05-10
Add alt tags to images

mjankowski:
Resolves https://github.com/thoughtbot/constable/issues/608

I feel like there's probably a better way to do the custom scrubber bit but not sure what.

### DIFF
--- a/lib/constable/markdown.ex
+++ b/lib/constable/markdown.ex
@@ -4,6 +4,6 @@ defmodule Constable.Markdown do
   def to_html(markdown) do
     markdown
     |> Earmark.as_html!(%Options{smartypants: false})
-    |> HtmlSanitizeEx.html5()
+    |> HtmlSanitizeEx.Scrubber.scrub(Constable.CustomScrubber)
   end
 end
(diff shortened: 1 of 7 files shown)


lib/constable/markdown.ex
```diff
@@ -1,9 +1,80 @@
 defmodule Constable.Markdown do
   alias Earmark.Options
 
+  defmodule CustomScrubber do
```
germsvel: Not sure if it's worth doing this or not, but what do you think about adding more tests to the `CustomScrubber` as to what it scrubs and what it doesn't? 

I know we have some tests, and I see the comment says that this is basically a copy of the `markdown_html.ex` scrubber, but might it be worth checking since we own the code now?

mjankowski: That's a good idea ... I will make a second pass at this where I at least *attempt* to borrow less from the other lib and just put our own rules (along with our own tests) in place, and see how it works out. If that turns out to be too clunky or feels like I'm just duplicating the test suite of the html_sanitizer_ex lib's various scrubbers, I'll back off and probably just keep something like this.

I also thought about putting the `CustomScrubber` in it's own file with an even bigger comment at the top about where it came from and what very specific changes I made, just to be totally clear.

PR #613, opened 2019-05-24
Adds "Your Announcements" to nav bar

germsvel:
What changed?
------------

This PR adds a new tab into the announcements-interests nav bar. The tab, titled "Your announcements", allows users to see only announcements they have made.

Why?
----

When users make announcements, they have the ability to edit them in the future. Users may also want to see comments people have left in one of their previous announcements. But I have always found it difficult to find my previous announcements.

Typically, I would open an issue before creating a PR to see if this is even wanted by others. But this is simple enough that I thought I'd just take a stab at it, and see if it makes sense to merge. If we don't think this will be useful (other than for me), I'm happy to close this PR.

Demo
----

add-your-announcements-link

Comments on some of the code
----------------------------

Most of the implementation was included alongside the implementation of the other two `announcement#index` actions, leveraging pattern matching to filter things correctly. It seems like there's a lot of logic in the controller, and it would be good to refactor it. I have decided not to do more of that in this PR to avoid introducing more changes.

There are also some unrelated changes in files I have touched due to the elixir formatter. I have decided to include those here rather than turning the formatter off, but it may be good to run the formatter across all files in a single PR so that we do not have constant PRs that have a lot of unrelated changes due to auto-formatting.

### DIFF
--- a/lib/constable/models/announcement.ex
+++ b/lib/constable/models/announcement.ex
@@ -39,32 +39,41 @@ defmodule Constable.Announcement do
   end
 
   def with_announcement_list_assocs(query \\ __MODULE__) do
-    from q in query, preload: [
-      :user,
-      interests: ^Interest.ordered_by_name,
-      comments: ^newest_comments_first(),
-      comments: :user
-    ]
+    from q in query,
+      preload: [
+        :user,
+        interests: ^Interest.ordered_by_name(),
+        comments: ^newest_comments_first(),
+        comments: :user
+      ]
+  end
+
+  def for_user(query \\ __MODULE__, user_id) do
+    query
+    |> where([a], a.user_id == ^user_id)
   end
 
   def search(query \\ __MODULE__, search_term, exclude_interests: excludes) do
     search_term = search_term |> prepare_for_tsquery
 
-    excludes = if Enum.empty?(excludes) do
-      [""]
-    else
-      excludes
-    end
+    excludes =
+      if Enum.empty?(excludes) do
+        [""]
+      else
+        excludes
+      end
 
     from(a in query,
       full_join: i in assoc(a, :interests),
       group_by: a.id,
       having: fragment("NOT ARRAY_AGG(?) @> ?", i.name, ^excludes),
-      where: fragment("to_tsvector('english', ?) || to_tsvector('english', ?) @@ plainto_tsquery('english', ?)",
-        a.title,
-        a.body,
-        ^search_term
-      )
+      where:
+        fragment(
+          "to_tsvector('english', ?) || to_tsvector('english', ?) @@ plainto_tsquery('english', ?)",
+          a.title,
+          a.body,
+          ^search_term
+        )
     )
   end
 
--- a/lib/constable_web/controllers/announcement_controller.ex
+++ b/lib/constable_web/controllers/announcement_controller.ex
@@ -13,12 +13,16 @@ defmodule ConstableWeb.AnnouncementController do
     index_page = all_announcements() |> Repo.paginate(params)
 
     conn
-    |> assign(:announcements, index_page.entries)
     |> assign(:show_all, true)
-    |> assign(:current_user, preload_interests(conn.assigns.current_user))
-    |> assign(:index_page, index_page)
-    |> page_title("Announcements")
-    |> render("index.html")
+    |> render_index(index_page)
+  end
+
+  def index(conn, %{"user_id" => user_id} = params) do
+    index_page = user_announcements(user_id) |> Repo.paginate(params)
+
+    conn
+    |> assign(:show_all, false)
+    |> render_index(index_page)
   end
 
   def index(conn, %{"interest" => interest, "page" => page}) do
@@ -27,11 +31,16 @@ defmodule ConstableWeb.AnnouncementController do
   end
 
   def index(conn, params) do
-    index_page = my_announcements(conn) |> Repo.paginate(params)
+    index_page = announcements_of_interest(conn) |> Repo.paginate(params)
 
     conn
-    |> assign(:announcements, index_page.entries)
     |> assign(:show_all, false)
+    |> render_index(index_page)
+  end
+
+  defp render_index(conn, index_page) do
+    conn
+    |> assign(:announcements, index_page.entries)
     |> assign(:current_user, preload_interests(conn.assigns.current_user))
     |> assign(:index_page, index_page)
     |> page_title("Announcements")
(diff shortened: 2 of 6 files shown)


lib/constable/models/announcement.ex
```diff
@@ -39,32 +39,41 @@ defmodule Constable.Announcement do
...
+    from q in query,
+      preload: [
+        :user,
+        interests: ^Interest.ordered_by_name(),
+        comments: ^newest_comments_first(),
+        comments: :user
+      ]
+  end
+
+  def for_user(query \\ __MODULE__, user_id) do
```
germsvel: This was a new function added to this file. The rest of the changes are due to auto-formatting.

lib/constable_web/controllers/announcement_controller.ex
```diff
@@ -27,11 +31,16 @@ defmodule ConstableWeb.AnnouncementController do
...
-    index_page = my_announcements(conn) |> Repo.paginate(params)
+    index_page = announcements_of_interest(conn) |> Repo.paginate(params)
 
     conn
-    |> assign(:announcements, index_page.entries)
     |> assign(:show_all, false)
+    |> render_index(index_page)
+  end
+
+  defp render_index(conn, index_page) do
```
germsvel: Decided to extract a lot of the common rendering logic from the `def index/2` functions in here.

lib/constable_web/views/announcement_view.ex
```diff
@@ -3,35 +3,33 @@ defmodule ConstableWeb.AnnouncementView do
...
   end
 
   def comma_separated_interest_names(_), do: ""
 
-  def class_for("all", %{params: %{"all" => "true"}}) do
-    "selected"
-  end
-  def class_for("mine", %{params: %{"all" => "true"}}), do: nil
-  def class_for("mine", _) do
-    "mine selected"
```
germsvel: There was no `css` for `mine`, so I just left `selected` in the changes. Also added another `class_for` for when we are clicking on `"your announcements"`.

test/controllers/announcement_controller_test.exs
```diff
@@ -11,19 +11,33 @@ defmodule ConstableWeb.AnnouncementControllerTest do
...
 
-    conn = get conn, Routes.announcement_path(conn, :index, all: true)
+    conn = get(conn, Routes.announcement_path(conn, :index, all: true))
 
     assert html_response(conn, :ok) =~ "Awesome"
     assert html_response(conn, :ok) =~ "Lame"
   end
 
-  test "#index only my announcements are shown by default", %{conn: conn, user: user} do
+  test "#index shows user's announcements when user_id is set", %{conn: conn, user: user} do
```
germsvel: This is the new test that was added. The rest of the changes in this file were done by the auto-formatter.

test/models/announcement_test.exs
```diff
@@ -75,15 +75,34 @@ defmodule Constable.AnnouncementTest do
...
+      announcement =
+        Announcement.with_announcement_list_assocs()
+        |> Repo.one()
+
+      assert announcement.interests |> List.first() == interest_a
+      assert announcement.interests |> List.last() == interest_b
+    end
+  end
+
+  describe ".for_user/1" do
```
germsvel: This is the new test in this file. The rest of the changes were done by the auto-formatter.

lib/constable_web/views/announcement_view.ex
```diff
@@ -3,35 +3,33 @@ defmodule ConstableWeb.AnnouncementView do
...
-    |> Enum.map(&(%{name: &1.name}))
-    |> Poison.encode!
+    |> Enum.map(&%{name: &1.name})
+    |> Poison.encode!()
   end
 
   def comma_separated_interest_names(interests) when is_list(interests) do
     interests
-    |> Enum.map(&(&1.name))
+    |> Enum.map(& &1.name)
```
MattMSumner: I'm assuming the parens changes are auto-formatting as well?

test/controllers/announcement_controller_test.exs
```diff
@@ -11,19 +11,33 @@ defmodule ConstableWeb.AnnouncementControllerTest do
...
+    insert(:announcement, title: "Awesome", user: user)
+    insert(:announcement, title: "Lame", user: other_user)
+
+    response =
+      conn
+      |> get(Routes.announcement_path(conn, :index, user_id: user.id))
+      |> html_response(:ok)
+
+    assert response =~ "Awesome"
+    refute response =~ "Lame"
```
MattMSumner: `Lame`

lib/constable_web/views/announcement_view.ex
germsvel: Yeap. I've turned off my auto-formatter in the past. I think I might just open a PR soon that auto-formats everything, and then this won't be a problem.

test/controllers/announcement_controller_test.exs
germsvel: I thought so too.

setphen: https://thoughtbot.slack.com/archives/C024HFHU9/p1558016515265000?thread_ts=1557955114.247600&cid=C024HFHU9

germsvel: I see. I'm guessing that means I should change this?

MattMSumner: Ha, I thought I was being funny but it looks like it would be good to change this, how would you feel about:

```suggestion
    refute response =~ "Should not appear"
```

germsvel: Updated this (and the test where I had copied the setup from) to avoid using the word "Lame" in fbae53a9b0e7320a0822391e1b996e081835845a. Thanks for the feedback!

lib/constable/models/announcement.ex
```diff
@@ -39,32 +39,41 @@ defmodule Constable.Announcement do
...
 
   def search(query \\ __MODULE__, search_term, exclude_interests: excludes) do
     search_term = search_term |> prepare_for_tsquery
 
-    excludes = if Enum.empty?(excludes) do
-      [""]
-    else
-      excludes
-    end
+    excludes =
```
lancejjohnson: Nitpick: What would you think about adding a function head to remove this conditional?

```elixir
def search(query \\ __MODULE__, search_term, exclude_interests: []) do
  search(query, search_term, exclude_interests: [""]
end

def search(query \\ __MODULE__, search_term, exclude_interests: [_|_] = excludes) do
   # ... the actual impl
end
```

germsvel: Hi Lance, thanks for this input! Lots of the changes in this PR were due to the formatter just fixing things like this one. I was trying not to introduce too many changes that would muddy the PR since the formatter did quite a few of these kinds of changes. So I'll hold off on making that change, though I like the suggestion you made.

PR #618, opened 2019-06-07
Make user names clickable to see their announcements

germsvel:
What changed?
============

In 1e357b6, we introduced a  `/announcements#index` route with the query param `user_id` in order to display a user's own announcements. We leverage the core of that functionality to display the list of announcements for _any_ user when we click their name.

The only strange thing about this is that when we select a user's name, we leave the nav bar in an unselected state (i.e. no tab is marked as "selected"), since seeing another user's announcements does not have an equivalent "tab" as we might have for "All announcements" or "Your announcements".

Why this change?
================

I often find that I expect to be able to navigate to a user's announcements by clicking on their name only to be surprised by the fact that I can't.

NOTE: I'm not sure if this is a good addition to Constable. I'd be interested in hearing other people's thoughts. It's possible that announcements are meant to be slightly more ephemeral and that we should just use the search functionality if we want to find old announcements. So I'm happy to close this if others find this unhelpful.

names-link-to-users-posts

### DIFF
--- a/lib/constable_web/templates/announcement/show.html.eex
+++ b/lib/constable_web/templates/announcement/show.html.eex
@@ -27,7 +27,7 @@
     <div class="media media-center">
       <img src="<%= gravatar @announcement.user %>" class="media-figure avatar-rounded avatar-rounded-large"/>
       <div class="media-body">
-        <h4 class="author"><%= @announcement.user.name %></h4>
+        <h4><%= link @announcement.user.name, to: Routes.announcement_path(@conn, :index, user_id: @announcement.user_id), class: "author" %></h4>
 
         <div class="announcement-metadata">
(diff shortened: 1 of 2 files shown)


germsvel:
@jutonz yeah, you bring up a good point about the route. Maybe I'm trying to fit a square peg into a round whole with this?

germsvel:
@jutonz I put the css class that was previously in the `h4` into the link so that it the author's link is consistent with the rest of the links in the page (which are all _not_ red but have the underline when hovering over them). Just wanted to send you an update since you'd reviewed this! 

Screen Shot 2019-06-14 at 9 58 53 AM

jutonz:
Sweet, thanks for the update @germsvel!

PR #623, opened 2019-06-19
Preview changes to new and edited markdown comments

### LINKED ISSUE #459: Preview markdown comments
When I'm writing a comment that includes some formatting and images. I want to be able to preview how it will render with one tap. That way I'm confident I won't have to edit it after it's posted.

drewvolz:
Resolves #459. A few ideas are accomplished in this PR:

**Toggling between composing/previewing markdown comments**

* The preview button hides the editable area and shows a markdown preview. 
* The write button hides the preview and presents the editable textarea.

**Saving and restoring drafted comments**

* Only for new comments –– not for edits
* WIP comments are saved to LocalStorage
* Drafts persist across tabs and page reloads
* They are not tied to posts or comments –– just the input fields

---

📸 Screenshots

~ | ~
--|--
Write | write
Preview | preview

### DIFF
--- a/lib/constable_web/templates/comment/edit.html.eex
+++ b/lib/constable_web/templates/comment/edit.html.eex
@@ -1,7 +1,7 @@
 <div class="container">
   <div class="comment-edit">
-    <button class="button-primary" id="comment-preview-text-button">Write</button>
-    <button class="button-primary" id="comment-preview-markdown-button">Preview</button>
+    <button class="tbds-button" id="comment-preview-text-button">Write</button>
+    <button class="tbds-button" id="comment-preview-markdown-button">Preview</button>
     <div class="comment-body" id="comment-preview"></div>
 
     <%= render "_form.html",
--- a/lib/constable_web/templates/comment/edit.html.eex
+++ b/lib/constable_web/templates/comment/edit.html.eex
@@ -1,5 +1,9 @@
 <div class="container">
   <div class="comment-edit">
+    <button class="button-primary" id="comment-preview-text-button">Write</button>
+    <button class="button-primary" id="comment-preview-markdown-button">Preview</button>
+    <div class="comment-body" id="comment-preview"></div>
+
     <%= render "_form.html",
       comment_changeset: @changeset,
       path: Routes.announcement_comment_path(@conn, :update, @announcement, @comment),
--- a/lib/constable_web/templates/comment/edit.html.eex
+++ b/lib/constable_web/templates/comment/edit.html.eex
@@ -1,8 +1,8 @@
 <div class="container">
   <div class="comment-edit">
-    <button class="tbds-button" id="comment-preview-text-button">Write</button>
-    <button class="tbds-button" id="comment-preview-markdown-button">Preview</button>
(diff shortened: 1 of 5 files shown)


tysongach:
@drewvolz I recently merged a few PRs that changed our CSS fairly heavily, and that has caused a conflict with this branch - sorry about that! Let me know if I can help with that…

drewvolz:
Thanks for the heads-up, @tysongach!

lib/constable_web/templates/comment/edit.html.eex
```diff
@@ -1,5 +1,8 @@
 <div class="container">
   <div class="comment-edit">
+    <button class="tbds-button comment-preview-text-button">Write</button>
+    <button class="tbds-button comment-preview-markdown-button">Preview</button>
```
germsvel: It seems we're rendering these buttons both on the `announcement/show.html.eex` and the `announcements/edit.html.eex` templates. Is this something that could go in the `_form.html` template since they both use that and since this functionality is related to that form?

germsvel: What do you think about changing the color of these buttons? They seem too much like primary buttons (that bright red) when I think they're more secondary. 

Screen Shot 2020-01-06 at 9 22 21 AM
Screen Shot 2020-01-06 at 9 22 30 AM

---

I added the `btn-secondary` class to them in a quick test, and now the problem is that their inactive state looks too much like it. 

Screen Shot 2020-01-06 at 9 38 36 AM

---

Then I tried a couple of other background colors from he `_variables.scss` file: 

Screen Shot 2020-01-06 at 9 38 58 AM

Screen Shot 2020-01-06 at 9 39 13 AM

---

I think I'm not the right person to figure this out 😄 but I do think it might make sense to not style them as the primary button (the "Post Comment" button). What do you think?

drewvolz: This is a wonderful suggestion. I'll refactor the buttons so that they live inside `comment/_form.html.eex`.

I've noticed some logic around submitting the form in `comment-form.js` and submitting, so I'll explicitly set these buttons to be of type "button" since most browsers set the default type of button as "submit".

drewvolz: Thanks for showing your thought process, that's really helpful. I believe you're correct that these buttons do not serve a primary purpose and perhaps should not sport the primary button styling.

I'll see what I can come up with, too, and maybe call upon a designer or two for their thoughts!

drewvolz: Do you have any thoughts around this, @tysongach?

drewvolz: A thought I have on the styling of these controls could involve
1. Subscribing the buttons to the `.tbds-button` + `.button-secondary` styles
2. Setting a border of zero on the inactive control

This would end up looking like:

button-secondary-animation

drewvolz: @smharley was kind enough to lend a hand! This looks **great**. Thank you, Steven!

image

PR #639, opened 2019-07-23
Underline links by default

entcheva:
Color differences in links can be hard/impossible to see for those with
color blindness. Let's underline links by default, and underline dashed
on focus & hover.

I'm having trouble getting Constable running locally (discussed in https://github.com/thoughtbot/constable/pull/603),
can someone confirm this works locally & doesn't look terrible?

### DIFF
--- a/assets/css/base/_typography.scss
+++ b/assets/css/base/_typography.scss
@@ -56,14 +56,14 @@ p {
 
 a {
   color: $action-color;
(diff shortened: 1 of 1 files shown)


entcheva:
Thanks for the screenshots, Drew! It's not beautiful, but it's functional so - merging.

PR #643, opened 2019-08-01
Base CustomScrubber from HTML5 scrubber

germsvel:
Resolves: https://github.com/thoughtbot/constable/issues/492

What changed?
============

We update the `Constable.CustomScrubber` to use the `HtmlSanitize.Scrubber.HTML5` module as a base instead of the `MarkdownHtml`.

Why?
====

Back in 7609c5338cadf99a0b232ba68e8243f2afab8768, we moved from using the html5 scrubber to using a modified version of the `MarkdownHtml` scrubber. We did this in order to allow `alt` attributes in `img` tags.

It turns out, however, that the `MarkdownHtml` module is far more restrictive than the html5 module we were using. And due to a bug in the `HTML5` scrubber, the `alt` attributes were being stripped form the `img` tag.

So we introduce a modified version of the HTML5 scrubber that fixes the bug with the `alt` attribute in `img` tags.


germsvel:
@pablobm I did see the warning, and thought it funny. But since we were already using it, and since it's an internal tool, I'm going with it. 😄 Thanks for the review!

PR #648, opened 2019-08-09
Introduce CSS architecture and organization, add stylelint

tysongach:
This organizes our CSS using the Inverted Triangle CSS methodology,
or ITCSS. It's high-level thinking behind the ordering is:

- Generic to explicit
- Low specificity to high specificity
- Far-reaching to localized

The layers consist of:

1. Settings - Global settings like variables; this should be
   non-rendering Sass (no CSS output).

2. Tools - Global tools like mixins and functions.

3. Generic - The first layer that produces CSS output, things like
   global box-sizing setup, resets/normalizers, etc.

4. Elements - Bare, unclassed HTML element styles, like `h1`,
   `ul`, and `a`.

5. Objects - Non-cosmetic design patterns like high-level containers
   and grid systems. If you're familiar with OOCSS (object-oriented
   CSS), the classic `media` object would go here. This layer is where
   class-based styling begins.

6. Components - Where the majority of styles will typically wind up,
   we're styling recognisable pieces of UI in this layer. This is still
   all class-based styles.

7. ~~Trumps~~ - Nope! We don't have these yet, but we should also use
   the term "utilities" instead of "trumps". This layer is utility and
   helper classes, hacks and overrides.

Where easily possible, this also moves a few classes around so that
each Sass partial contains one single key selector, not multiple.

Read more about ITCSS:
https://www.creativebloq.com/web-design/manage-large-css-projects-itcss-101517528

Co-authored-by: Keiran King <keiran@thoughtbot.com>

---

This PR also add stylelint for linting our CSS/SCSS and I've enabled Hound to run it.


tysongach:
I marked all of Hound’s comments as “resolved” because—beyond the run of stylelint's autofix feature—this PR is just about organizing and getting the tooling in place. There were valid linting warnings raised, but we can fix them via refactoring in future PRs.

PR #649, opened 2019-08-12
Remove unused neat dependency

tysongach:
We stopped using Neat in https://github.com/thoughtbot/constable/commit/438cfbbf13f8f147649676037244b52d0d351be9

We need to add `node-sass` as a dependency because we require it, but
we were implicitly using it via `neat`'s dependency on it.


tysongach:
It looks like the test might be failing because we have an implicit dependency on `node-sass`. `neat` v1 had it as a dependency, so we were implicitly consuming that without declaring it ourselves.

Issue #652, opened 2019-08-21
Floating current nav item indicator on mobile


cpytel:
The current navigation item is indicated by a dark line underneath it. This looks find on desktop-width browsers, but at mobile width, this indicator is floating as the navigation items don't appear all on one line.

![IMG_0552](https://user-images.githubusercontent.com/5015/63426773-a41e0700-c3e1-11e9-8590-8967f590f87e.jpeg)

PR #697, opened 2019-09-04
Fix secondary nav layout on small screens

tysongach:
Fixes https://github.com/thoughtbot/constable/issues/652

On small viewports, the nav will now scroll horizontally:

Screen Shot 2019-09-04 at 09 01 49
Screen Shot 2019-09-04 at 09 02 14


assets/css/components/_segmented-control.scss
```diff
@@ -0,0 +1,26 @@
...
+    @include padding(null ($base-spacing - $small-spacing));
+  }
+
+  a {
+    display: inline-block;
+    padding: $small-spacing;
+    white-space: nowrap;
+
+    &.selected {
+      box-shadow: inset 0 -2px $dark-gray;
```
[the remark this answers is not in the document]
tysongach: I agree with this, but left it because it will require a good amount of refactoring because our `class_for` helper doesn’t allow us to pass in arbitrary classes to go alongside the `selected` class. So this nesting is needed…

assets/css/components/_segmented-control.scss
```diff
@@ -0,0 +1,26 @@
+.segmented-control {
```
vendelal: I see that "segmented-control" is a term that was already being used, but it feels a bit clunky to me. `tabs` is more descriptive IMO

tysongach: Yep, I could go either way.

I believe the term “Segmented Control” comes from [Apple’s HIG](https://developer.apple.com/design/human-interface-guidelines/ios/controls/segmented-controls/). Perhaps that’s why it was used here?

tysongach: Frankly, I want to remove this entire bar eventually…I think there are better design patterns we could leverage to filter the messages (which is really all this is doing, plus a link to Interests).

vendelal: Ooh yeah that makes sense about Apple. I had a similar thought when I went to look at the live site. There's a lot of blank space not being used 🤠 

I'm not super opinionated about this, so if you leave it as `segmented-control` I won't fight you.

PR #726, opened 2019-09-13
Improve comments area empty state

### LINKED ISSUE #682: Improve empty state when an announcement has no comments
When viewing an announcement that has no comments, the “Comments” heading has nothing under it and the call-to-action (to leave a comment) feels baren and uninviting:

<hr>

Screen Shot 2019-08-28 at 09 11 43

WilHall:
This PR fixes #682 by improving the comments empty state.

I would love feedback on what message we would like to display here.

Image 2019-09-12 at 9 37 36 PM

### DIFF
--- a/assets/css/components/_comments.scss
+++ b/assets/css/components/_comments.scss
@@ -1,6 +1,6 @@
 .comments-placeholder {
-  display: none;
   color: #999;
+  display: none;
   font-size: 1.3rem;
   text-align: center;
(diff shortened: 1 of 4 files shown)


WilHall:
@stratigos Hm, yeah I similarly came up with some long messages at first but then shortened to what I currently have. I'll leave it for now and we can always iterate 🙂

Issue #683, opened 2019-08-28
Using Gravatar for avatars


tysongach:
We’re currently using [Gravatar](https://en.gravatar.com/) for avatar images. There are two issues I see with that:

1. It means I have to use yet another service to manage a photo of me. Some folks might not even know what Gravatar is, and it’s a bit weird to describe (I think this leads to point 2 below).
1. It’s not uncommon to see this on Constable instead of a colleagues face:

   ![grav](https://user-images.githubusercontent.com/903327/63859436-5a9b6200-c975-11e9-980a-a2408151f0c5.jpg)

I have a hunch that Gravatar is being used because it’s implementation on the dev-side is fairly straightforward.

Would it be possible and/or valuable to leverage the avatars we already have in Hub or Slack?

WilHall:
I agree 💯 that for thoughtbot we should ideally integrate with hub. Did you have thoughts on how that might look alongside an open-source compatible offering (probably Gravatar)?

mjankowski:
I suspect that the lib we are using (ExGravatar) might not support this and we'd have to roll our own or use something else -- but in theory we could leave the hostname portion configurable, and support a "gravatar-style" URL from hub itself, if we care deeply about this.

So our TB instance would hit something like `https://hub.thoughtbot.com/avatars/:md5-email-hash`, but the open source version would default that configuration to stock gravatar urls.

WilHall:
This was fixed by #743

Issue #460, opened 2018-04-05
Show users’ studio


mborsare:
When a person I don't know posts a great announcement, I want to know which office they work out of so I can learn a little more about them.

stratigos:
Would a link to their hub page satisfy this?

I agree that its useful to know which studio theyre working from, but then I wonder what other information folks would want about the announcer. That led me to thinking a link to their hub page would reveal just about anything of interest, though I then see the issue of needing an additional step to get at info (click link, view new page).

tysongach:
I made quick mockup using a badge to show the studio people work in:

<img width="666" alt="Screen Shot 2019-09-28 at 09 53 17" src="https://user-images.githubusercontent.com/903327/65817632-c0c70f00-e1d7-11e9-974e-ab7606fc3f0c.png">

My thinking is that the badges would only exist on announcement show pages (not index pages) at the top for the announcement’s author, and also in the comments section (as shown above). Maybe the colors are too much, or not needed.

---

> I agree that its useful to know which studio theyre working from, but then I wonder what other information folks would want about the announcer.

Yeah, perhaps job title would be valuable info, as well?

---

In https://github.com/thoughtbot/constable/commit/317bf34db524ec56358d171afbc5f8972729f129 we made it possible to view announcements from a specific user. For example, here is an index of Stephanie’s announcements:

<img width="1071" alt="Screen Shot 2019-09-28 at 10 00 47" src="https://user-images.githubusercontent.com/903327/65817570-e1429980-e1d6-11e9-9355-84697b945ca4.png">

A page title/heading could be really useful on this page (“Stephanie Kuroda’s announcements”). That might pave the way to have a little header section, with the person’s name, job title and studio location, and perhaps a link to their Hub profile.

WilHall:
@tysongach I like this a lot! It was nothing like what I was imagining, but I think it provides a lot more value than other options because it's _right there_ and the colors make it really easy to glance and see the office info!

brimatteng:
With the thoughtbot move to remote-first and no offices I'm not sure this really makes sense anymore. I'm going to close this issue

PR #729, opened 2019-09-14
Add Hub integration

### LINKED ISSUE #728: Hub integration
Several open issues (#460, #683) would require Hub integration to resolve.

This issue is for the back-end hub integration, and will simply add a link to the user's profile in Hub to the constable user settings page.

After the Hub integration is in place, future issues can utilize it.

WilHall:
Fixes #728.

This PR:
 - Adds two new environment variables `HUB_URL` and `HUB_API_TOKEN`
 - Adds `neuron` as a dependency for querying Hub's GraphQL endpoint
 - Adds a `HubProfile` service object for fetching user profile information from hub
 - Adds a `View your Hub profile` link in the user settings view

There are other interesting features that can be implemented once the hub integration is in place including [using Hub profile images instead of Gravatar](https://github.com/thoughtbot/constable/issues/683), [showing a user's office location](https://github.com/thoughtbot/constable/issues/460), and some other ideas mentioned in thoughtbot/hub#2583


WilHall:
> What’s the user value in this?

Out thought was that as we add more hub integrations, it will need to be evident where the user can go to change their profile information. The link is merely a convenience, but I think ideally we also want to convey that some of your profile information in Constable is editable in Hub.

Do you have any ideas of how we can convey that, with or without the link? Maybe some help text in the settings form?

tysongach:
> Out thought was that as we add more hub integrations, it will need to be evident where the user can go to change their profile information.

This makes a ton of sense and provides specific user value (thanks for clarifying!). You could write a Job-to-be-Done story around that. But I’m not sure a “View your Hub profile” link makes sense right now, given that no user information is being pulled from Hub yet.

WilHall:
@tysongach Makes sense - I suppose we could adjust this PR to have no visible user change and just add the support for Hub integration, and then discuss specific integrations in additional stories. That would be fine with me!

WilHall:
I have removed the hub profile link, but left the `HubProfile` service in place.

.sample.env
```diff
@@ -6,3 +6,5 @@ OUTBOUND_EMAIL_DOMAIN=staging.constable.io
 PERMITTED_EMAIL_DOMAIN=thoughtbot.com
 SHUBOX_SCRIPT_URL=get-from-shubox-dashboard
 SLACK_WEBHOOK_URL=https://fakeslack/webhook
+HUB_URL=https://staging-hub.thoughtbot.com
+HUB_API_TOKEN=get-value-from-hub-staging
```
stratigos: What do you think about adding the env param name to this placeholder value?

E.g., `get-API_TOKEN-value-from-thoughtbot-hub-staging`

I wasnt sure what value this needed to be when I first glanced at this file. I would intuitively guess that Hub would issue Constable its own API key, so that Id need to look in Constable-staging, rather than Hub-staging. I then looked at the Heroku settings page for Hub-staging, and it felt obvious to me that the value for `API_TOKEN` is what I need to select. Hub also has a lot of other env params. Im wondering if anyone else would pause for a moment over this like I did.

test/services/hub_profile_test.exs
```diff
@@ -0,0 +1,57 @@
...
+
+  import Mock
+  alias Constable.Services.HubProfile
+
+  describe "#profile_url" do
+    test "returns the hub profile url for a user" do
+      email = "test@example.com"
+      user = insert(:user, email: email)
+
+      mock_slug = "mollusk"
```
stratigos: :laughing:

PR #730, opened 2019-09-14
Fix slack channel form button overlap

### LINKED ISSUE #680: Slack channel buttons overlap
When editing a Slack channel for an interest using a device with a small screen, the “Remove Channel” button can overlap the “Save Slack Channel” button:

IMG_3339

WilHall:
Fixes #680 (using `tbds`)!

This PR:
 - Fixes the slack channel button overlap
 - Adds `tbds` classes and markup to the slack channel form

## Before
Image 2019-09-14 at 11 15 31 AM

## After
Image 2019-09-14 at 11 15 31 AM


WilHall:
@tysongach My first thought here was to use flexbox to resize and wrap the buttons, but I noticed there wasn't a helper in `tbds` for this like the stack classes. I'm curious if that's intentional, and if you'd suggest implementing the flex for the buttons here in Constable or if that would be a candidate to add to `tbds?

tysongach:
Great question. tbds Stacks do not support wrapping, as of now. Stacks are a high-level object, though, and it’s looking like this UI might require a more specific solution. For example, wrapping the buttons onto multiple lines could be good, but we’ll then need whitespace between the buttons:

Screen Shot 2019-09-17 at 10 08 10

We may also want to make the buttons full-width, as well, otherwise the alignment looks funny:

Screen Shot 2019-09-17 at 10 08 39

So, we may be shooting for something like this:

Screen Shot 2019-09-17 at 10 10 07

Given the contextual wrapping, spacing, and full-width buttons, this is sounding like a new “thing” and not within the scope of tbds Stacks.

Is that helpful?

WilHall:
@tysongach agree - i'll keep this PR open until we merge #737 and then apply some flex styling on top of that.

WilHall:
@tysongach I'm curious what you think of the flexbox approach in my latest commit. It is much more adaptive, though the buttons are not quite full-width (because of margins) but I think this makes the layout look better on all sizes.

assets/css/components/_buttons.scss
```diff
@@ -0,0 +1,11 @@
+.flex-button-container {
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+  justify-content: space-between;
+
+  > * {
```
tysongach: It’d be great to avoid wildcard selectors like this, in favor of applying CSS classes directly to the items we want to style. These types of selectors feel concise, but get unwieldy quickly.

assets/css/app.scss
```diff
@@ -27,6 +27,7 @@
 @import "components/app-header";
 @import "components/at-mention";
 @import "components/box";
+@import "components/buttons";
```
tysongach: Since we are not styling buttons in this partial, how do you feel about naming the partial `flex-button-container`?

lib/constable_web/templates/slack_channel/edit.html.eex
```diff
@@ -22,25 +22,24 @@
...
             gettext("Save Slack Channel"),
             id: "submit-channel-name",
             class: "tbds-button"
           ) %>
+          <%= if @interest.slack_channel do %>
+            <%= link(
+              to: Routes.interest_slack_channel_path(@conn, :delete, @interest),
+              method: :delete,
+              data: [confirm: gettext("Are you sure?"), role: "remove-channel"],
+              class: "tbds-button button-secondary remove-slack-channel"
```
tysongach: It looks like we removed the `remove-slack-channel` CSS above - can this class be removed here?

```suggestion
              class: "tbds-button button-secondary"
```

Issue #700, opened 2019-09-04
Settings page is broken


tysongach:
At the bottom of announcement subscription emails, there’s a “Manage your notification settings” link:

<img width="669" alt="Screen Shot 2019-09-04 at 17 45 35" src="https://user-images.githubusercontent.com/903327/64294178-3541c180-cf3c-11e9-9472-dd2df61bb956.png">

When I follow that link, I'm taken to `https://constable.io/settings`, but the page appears broken:

<img width="566" alt="Screen Shot 2019-09-04 at 17 45 51" src="https://user-images.githubusercontent.com/903327/64294262-54d8ea00-cf3c-11e9-93c5-3f6589cb3eef.png">

tysongach:
Two related PRs:

- Settings page added: https://github.com/thoughtbot/constable/pull/196
- Settings moved into a modal: https://github.com/thoughtbot/constable/pull/378

I’ve been thinking about opening an issue about moving the Settings out of the modal and onto a page with a URL. I didn’t realize that `/settings` sort of exists already until I hit this bug just now.

I’m not sure the modal is really needed, and it modals have a tendency to introduce serious accessibility issues. So, given this issue, I’m inclined to get `/settings` working again, and perhaps remove the modal (replacing it with a link to `/settings`).

tysongach:
@WilHall Do you have some time to take a look at fixing this?

tysongach:
Never mind, I think I just found the culprit:

https://github.com/thoughtbot/constable/blob/60e6531ba5cb5e83dd4868bbac691a5ae523db2e/lib/constable_web/controllers/settings_controller.ex#L8

tysongach:
I’m going to investigate removing the modal/dialog UI from the settings and instead use a standard page view. Doing so would probably fix this issue, as well.

WilHall:
@tysongach Happy to help out with this later this week if it's still open. I think removing the modal and fixing the settings page is the right path for this.

PR #738, opened 2019-09-18
Fix Settings page, remove Settings modal

tysongach:
This fixes the rendering of the Settings page, which was not rendered
using a layout and therefore lacked CSS styling.

In conjunction, we remove the modal UI from the viewing of Settings and
instead leverage a standard page view. Modals often pose accessibility
issues (ours did) and it meant that did not have a URL to point at for
settings (our emails point to a `/settings` URL, though).

Fixes: https://github.com/thoughtbot/constable/issues/700

## Before

**Viewing `/settings`**

Screen Shot 2019-09-18 at 11 50 42

**Viewing settings modal**

Screen Shot 2019-09-18 at 11 50 54

## After

Screen Shot 2019-09-18 at 11 45 20


tysongach:
Aw, thanks, Todd! I appreciate the review.

Issue #733, opened 2019-09-17
Pull name from hub instead of duplicating in Constable


WilHall:
Mentioned by @tysongach in [a comment](https://github.com/thoughtbot/constable/issues/728#issuecomment-531783032) on #728.

A great follow-up to hub integration would be to pull user names from hub so that we are not storing that duplicate information in Constable.

PR #743, opened 2019-09-19
Remove gravatar, use Hub profile images

WilHall:
Fixes #568 

## This PR
 - Removes gravatar
 - Sources user profile images from Hub
 - Adds text to the settings page explaining where the photo comes from
 - Adds a default profile image of Ralph if the user does not have a profile image.

## Screenshots

<details>
    <summary>Avatars</summary>

    
Image 2019-09-19 at 5 51 43 PM

</details>

<details>
    <summary>Settings page</summary>

    
Image 2019-09-19 at 5 51 49 PM

</details>


lib/constable_web.ex
```diff
@@ -76,14 +75,12 @@ defmodule ConstableWeb do
       use Ecto.Schema
       import Ecto.Changeset
       import Ecto.Query
-      import Exgravatar
       @timestamps_opts [type: :utc_datetime, usec: false]
     end
   end
 
   def serializer do
```
stratigos: What would you think of removing `serializer` entirely? Please pardon my naïvety here, as this may be a required Phoenix protocol or something of that nature.

It looks like this is only in use in [User serializer](https://github.com/thoughtbot/constable/blob/620e583434b2dde2d0784adfdf7f65b64bf46e93/lib/constable/models/serializers/user.ex). From what I gather, it would simply unquote here and do nothing. Is it O.K. to also delete the line: `use ConstableWeb, :serializer` from the `User` serializer?

WilHall: I was actually hoping someone with more Phoenix experience would chime in! When I was looking at this I thought it should be fine to just remove it from here and remove the use in the user serializer, but I wasn't sure if there was a "right" thing to do in Phoenix. I'm inclined to remove it if we both think that's better.

WilHall:
> I think something here has caused a callback used by many tests to fail, though Im unsure what it is.

I noticed that, but wanted to get this PR up for code review. I haven't dug too far into the issue but the stack traces appear to be when unrelated tests call through the `HubProfile` service and that results in an HTTP call - which is not what we want in the tests.

I need to figure out the best way to mock the service in these cases, without explicitly mocking it in the unrelated tests. @stratigos Do you have any thoughts on that?

WilHall:
Turns out the `FakeGoogleStrategy` was a perfect example of mocking a service out for all tests using `Pact`.

lib/constable/services/hub_profile_provider.ex
```diff
@@ -0,0 +1,53 @@
+defmodule Constable.Services.HubProfileProvider do
+  alias Constable.Services.ProfileProvider
+  @behaviour ProfileProvider
+  use Memoize
+
+  @default_image_url "/images/ralph.png"
```
tysongach: I don’t see a `ralph.png` in the [`images` directory here in Constable](https://github.com/thoughtbot/constable/tree/master/assets/static/images)…is this coming from Hub?

lib/constable_web.ex
stratigos: Well it looks like the recent change introduced `profile_provider/0`, so now `serializer/0` is here to stay :moyai:

lib/constable/services/hub_profile_provider.ex
WilHall: It was added to the `images` directory [as part of this PR](https://github.com/thoughtbot/constable/pull/743/files#diff-4c5e766b6bbf2c9c11e874e97076009f)

tysongach: Whoops! Thanks for clarifying. I sometimes miss those in GitHub diffs because they are collapsed.

WilHall:
> Does this move to Hub impact the open-source nature of this project to the extent at which we should communicate something, say in the README? Or is that not necessary?

@tysongach Adding something to the `README` makes sense for this PR, but have we determined a stance yet?

I thought your comment on #683 summed it up well:
> I don’t. But I’m wondering: what is the intent of having Constable open-source? Is it to share our Elixir work, or is it to actually be used by others?

With the way this particular PR is implemented it would definitely be feasible to write a gravatar [profile provider](lib/constable/services/profile_provider.ex) thus maintaining the open source option, but I feel like some of the features we want to implement (#733, #460, likely others) beg for a tighter hub integration and I think in general I am in favor of that.

What are your thoughts on that?

tysongach:
@WilHall I wanted to prompt the questions, but would lean on German to answer them, [as Constable’s Product Owner](https://hub.thoughtbot.com/products/constable). He’s out until mid-October, though. Maybe we move forward with this change and merge, but hold off on other Hub integrations for a minute? As far as I know, they are small life improvements, but nothing too major.

WilHall:
@tysongach Sounds good to me. Thanks for discussing 🙂

PR #745, opened 2019-09-20
Reduce layout duplication

### LINKED ISSUE #739: Reduce duplication of layouts
We have two layouts which are nearly identical:

1. [`layout/app.html.eex`](https://github.com/thoughtbot/constable/blob/master/lib/constable_web/templates/layout/app.html.eex)
1. [`layout/login.html.eex`](https://github.com/thoughtbot/constable/blob/master/lib/constable_web/templates/layout/login.html.eex)

It seems like the primary difference is that `app` includes the primary header/nav as well as a few `<script>`s in the `<head>` and `login` doesn’t.

Can these be combined?

WilHall:
Fixes #739 - combines the app and login layouts to reduce duplication.

This PR should not have any visible effect on page layout.


lib/constable_web/templates/session/new.html.eex
```diff
@@ -1,4 +1,4 @@
-<div class="container container-x-small">
+<div class="container container-x-small container-full-height tbds-app-frame__body--center-items">
```
tysongach: The CSS class `tbds-app-frame__body--center-items` (a modifier in BE**M**) should be applied only alongside it’s “element” (`tbds-app-frame__body`). Otherwise, we’re no longer _modifying_ the element, we’re just leveraging this particular CSS class to apply some styles.

I _think_ this separation of element<>modifier may have also created the need for the `container-full-height` class, because the centering in `tbds-app-frame__body--center-items` doesn’t work if `container` doesn’t have `height`.

All that to say: it seems like we have a lot of implicit connections between our CSS classes here and it’d be great to tease them apart.

I’m guessing the thing that led to this is that we want the vertical centering (`--center-items`) on the login screen, but nowhere else. If that’s the case, I see two potential options forward (open to other ideas, too!):

1. Move `<main>` out of the layout and into each of the template files. Perhaps a bit repetetive, but it is just one element; not the worst problem.
1. Somehow pass a variable upward from the `session/new` template to the layout and use it to opt-in to the `--center-items` modifier class only for the screen.

WilHall: I went with a different approach that I think is more Phoenix-y: Move the definition of the `main` element into a partial called `_content_container.html.eex` and created a view helper `content_container` which `app.html.eex` uses to render it. It will render **either** `_content_container.html.eex` defined by the view if it exists, or the default one in `layout/_content_container.html.eex`.

This is more flexible as it doesn't cause any unnecessary duplication or codify the specific difference in a variable we'd have to pass.

PR #748, opened 2019-09-21
Adjust viewport and content padding

### LINKED ISSUE #679: Elements sometimes placed on the bottom edge of the screen
On some devices (like on an iPhone X, seen below), elements are placed right on the bottom edge of the screen. This can make them hard to read, or interact with.

IMG_3341

<br><hr><br>

IMG_3342

WilHall:
Fixes #679

## This PR
 - Sets the viewport meta tag `viewport-fit` to `cover` so that devices with "safe areas" display content in the safe areas
 - Adds a base padding to the `main` content area
 - Adds "safe inset" padding to `main` and `header` so that page content stays visible outside of safe areas

## Screenshots

<details>
    <summary>Portrait Top</summary>

    
Image 2019-09-21 at 3 18 12 PM

</details>

<details>
    <summary>Portrait Bottom</summary>

    
Image 2019-09-21 at 3 18 16 PM

</details>

<details>
    <summary>Landscape top</summary>

    
Image 2019-09-21 at 3 18 23 PM

</details>

<details>
    <summary>Landscape bottom</summary>

    
Image 2019-09-21 at 3 18 27 PM

</details>

### DIFF
--- a/assets/css/elements/_layout.scss
+++ b/assets/css/elements/_layout.scss
@@ -1,16 +1,22 @@
 html {
   box-sizing: border-box;
+  height: 100%;
+  width: 100%;
+}
+
+body {
+  min-height: 100%;
+}
+
+main {
+  padding: 0
+    calc(#{$tiniest-spacing} + env(safe-area-inset-right))
+    calc(#{$tiniest-spacing} + env(safe-area-inset-bottom))
+    calc(#{$tiniest-spacing} + env(safe-area-inset-left));
 }
 
 *,
 *::before,
 *::after {
   box-sizing: inherit;
-}
-
-html,
-body {
-  height: 100%;
-  min-height: 100vh;
-  width: 100%;
 }
--- a/assets/css/elements/_layout.scss
+++ b/assets/css/elements/_layout.scss
@@ -1,11 +1,5 @@
 html {
   box-sizing: border-box;
-  height: 100%;
-  width: 100%;
-}
-
-body {
-  min-height: 100%;
 }
 
 main {
--- a/assets/css/elements/_layout.scss
+++ b/assets/css/elements/_layout.scss
@@ -2,13 +2,6 @@ html {
   box-sizing: border-box;
 }
 
-main {
-  padding: 0
-    calc(#{$tiniest-spacing} + env(safe-area-inset-right))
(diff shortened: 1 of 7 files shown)


lib/constable_web/templates/layout/app.html.eex
```diff
@@ -3,7 +3,7 @@
   <head>
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
-    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta name="viewport" content="width=device-width, initial-scale=1 viewport-fit=cover">
```
tysongach: Is a comma needed here?

```suggestion
    <meta name="viewport" content="width=device-width, initial-scale=1, viewport-fit=cover">
```

assets/css/elements/_layout.scss
```diff
@@ -1,16 +1,22 @@
 html {
   box-sizing: border-box;
+  height: 100%;
+  width: 100%;
```
tysongach: I know this was existing and you just moved it up here, but I don’t _think_ `height` and `width` are needed and can be ✂️

assets/css/elements/_layout.scss
```diff
@@ -1,16 +1,22 @@
...
+}
+
+body {
+  min-height: 100%;
```
tysongach: I don’t _think_ this is needed either.

assets/css/elements/_layout.scss
```diff
@@ -1,16 +1,22 @@
...
+}
+
+main {
```
tysongach: It’d be great to move this to a class (like `app-main` or something) and into `components/`. `body` and `html` are required elements in any HTML document, but `main` is something we added and it’s contextual to the individual screens. So I like using classes in that case over far-reaching element styles. I have no idea if I just made any sense.

Perhaps an example would be useful: there are cases where two `main` elements are allowed. Having a dialog/modal is one…when the dialog is _shown_, it’s `main` is considered the primary content and the screen underneath should hide (`hidden`) it’s `main` element. So this is a case where we would probably not want these two `main` elements styled the same and having a class allows you to apply styles more contextually.

WilHall: Yes, that makes sense! :+1:

PR #751, opened 2019-09-22
Add mix test watch as a dev dependency

WilHall:
This PR adds [mix-test.watch](https://github.com/lpil/mix-test.watch) as a dependency.

This makes it possible to run tests when files change (`mix test.watch` for all tests or `mix test.watch --stale` for only tests related to changed files). This makes TDDing more streamlined.


germsvel:
I'm gonna go ahead and close this since it's not really needed, and it's a stale PR. If anyone wants to reintroduce this, please let me know. We can reopen it.

PR #767, opened 2019-09-28
Refactor media queries and breakpoints

tysongach:
In [438cfbb] we removed Neat (v1), but copied over its `media` mixin
because, at the time, it was used quite a bit and would've been more
complicated to pull out. That situation has changed and we're now in a
better position to remove it.

The `media` mixin we copied over is not a great mixin because it uses
string parsing instead of arguments. This has perhaps led us to use a
mix of `min-width` and `max-width` queries. It made our Sass overly
complex and hard to follow.

This commit:

- Removes the `media` mixin in favor of using vanilla CSS
  `@media` queries
- Moves everything over to `min-width` queries (commonly referred to as
  "mobile-first")

[438cfbb]: https://github.com/thoughtbot/constable/commit/438cfbbf13f8f147649676037244b52d0d351be9

### DIFF
--- a/assets/css/_shims.scss
+++ b/assets/css/_shims.scss
@@ -20,12 +20,10 @@
   }
 }
 
-@include media($medium-screen) {
-  .app-header {
-    .tbds-button.app-header__announce-button {
-      .tbds-button__icon--start {
-        margin-inline-end: 0;
-      }
+.app-header__announce-button {
+  .tbds-button__icon--start {
+    @media (max-width: $breakpoint-2) {
+      margin-inline-end: 0;
     }
   }
 }
--- a/assets/css/app.scss
+++ b/assets/css/app.scss
@@ -2,10 +2,7 @@
 
 @import "~@thoughtbot/design-system/src/index";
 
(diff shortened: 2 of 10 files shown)


assets/css/_shims.scss
```diff
@@ -20,12 +20,10 @@
...
 
-@include media($medium-screen) {
-  .app-header {
-    .tbds-button.app-header__announce-button {
-      .tbds-button__icon--start {
-        margin-inline-end: 0;
-      }
+.app-header__announce-button {
+  .tbds-button__icon--start {
+    @media (max-width: $breakpoint-2) {
```
sneamonitakis: Did we want to keep this as `max-width` for any reason?

tysongach: Yeah, this is a specific case. It’s in the `_shims.scss` partial and is fixing a regression the design system is causing. All styles in this partial should be temporary fixes and be removed once the real fix is in place.

Basically, we need to rethink Constable's “announce” button on small screens, or the design system needs to allow icon-only buttons at specific viewports.

PR #771, opened 2019-10-03
Disallow creation of announcements with no interests

### LINKED ISSUE #740: Posting an announcement to/with zero Interests
It is currently possible to post an announcement to/with zero Interests. [A recently posted announcement](https://constable.io/announcements/2218-new-advisor-skill-added-to-hub) made me aware of that.

At least in the case of this particular announcement, it being “tagged” with no Interest means the information is likely to have not reached people whom might find it valuable to have been notified of its existance.

I spoke with @knittingarch, who posted the announcement, to learn more about why no Interests were added, to which they very kindly responded with:

> I didn't realize I needed to add one.  I assumed since this applied to everyone, I shouldn't add anything.

This feels like a breakdown in the design and begs the question: How might we make it more clear how Interests work and/or when an announcement will (potentially) not reach anyone?

---

Interestingly, we already have a feature that will tell you how many people subscribe to the Interests you’ve selected:

Screen Shot 2019-09-19 at 14 43 13

…but that’s only if you add at least one Interest…

WilHall:
Fixes #740. 

The PR disallows the creation of announcements with no interests by adding a validation before the announcement is created, and returning an error that is displayed in the form.

Image 2019-10-02 at 9 02 44 PM

### DIFF
--- a/lib/constable/services/announcement_creator.ex
+++ b/lib/constable/services/announcement_creator.ex
@@ -13,17 +13,29 @@ defmodule Constable.Services.AnnouncementCreator do
 
   def create(params, interest_names) do
     changeset = Announcement.create_changeset(%Announcement{}, params)
-    case Repo.insert(changeset) do
-      {:ok, announcement} ->
-        announcement
-        |> add_interests(interest_names)
-        |> Repo.preload(:user)
-        |> subscribe_author
-        |> email_and_subscribe_users
-        |> SlackHook.new_announcement
 
-        {:ok, announcement}
-      error -> error
+    case interest_names do
+      [] ->
+        {:error,
+         changeset
+         |> Ecto.Changeset.add_error(:interests, "you must select at least one interest")
+         |> Map.put(:action, :update)}
+
+      _ ->
+        case Repo.insert(changeset) do
+          {:ok, announcement} ->
+            announcement
+            |> add_interests(interest_names)
+            |> Repo.preload(:user)
+            |> subscribe_author
+            |> email_and_subscribe_users
+            |> SlackHook.new_announcement()
+
+            {:ok, announcement}
+
+          error ->
+            error
+        end
     end
   end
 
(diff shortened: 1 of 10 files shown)


lib/constable/services/announcement_creator.ex
```diff
@@ -13,17 +13,29 @@ defmodule Constable.Services.AnnouncementCreator do
...
+            |> Repo.preload(:user)
+            |> subscribe_author
+            |> email_and_subscribe_users
+            |> SlackHook.new_announcement()
+
+            {:ok, announcement}
+
+          error ->
+            error
+        end
```
stratigos: I am having some kind of reaction to the nested `case` statement, but I cant really describe it. :thinking: I actually dont see anything wrong with it, but its triggering some kind of fear of future complexity if this direction continues. 

Im also not super experienced with the general culture of Elixir, and am wondering, are nested `case` statements typical? I did a web search and haphazardly clicked on this [SO qustion](https://stackoverflow.com/questions/50827551/elixir-how-to-avoid-deeply-nested-case-statements), which introduced me to the [`with` function](https://hexdocs.pm/elixir/Kernel.SpecialForms.html#with/1):

```elixir
opts = %{width: 10}
with {:ok, width} <- Map.fetch(opts, :width),
     {:ok, height} <- Map.fetch(opts, :height) do
  {:ok, width * height}
else
  :error -> {:error, :wrong_data} # fail case reached, since no `opts[:height]`
end
```

I suppose something like that :point_up_2: would be how to deal with future complexity, if someone feels included to add more `case` nesting. 

Im not suggesting a change now, as I cant tell if its just adding unecessary complexity, and I do think this code is fine as-is. But I can share at least that I learned a new pattern from this :grin:, and am wondering if anyone else has any thoughts on nested `case` statements in elixir :alembic:

lib/constable/services/announcement_creator.ex
```diff
@@ -49,27 +61,27 @@ defmodule Constable.Services.AnnouncementCreator do
...
 
-    Emails.new_announcement_mention(announcement, users) |> Mailer.deliver_later
+    Emails.new_announcement_mention(announcement, users) |> Mailer.deliver_later()
     announcement
   end
 
-  defp filter_author(author_id, users)  do
-    Enum.reject(users, fn(user) ->
+  defp filter_author(author_id, users) do
+    Enum.reject(users, fn user ->
```
stratigos: Im curious, did the formatter change this line? (removing parens) Im just trying to learn about its preferences regarding paren use :)

WilHall: Yes, there are several formatter changes in this PR. I'm not familiar with our formatter configuration, so I have just been letting it do its thing. It's something I'd love to discuss with @germsvel when he is back.

lib/constable/services/announcement_creator.ex
WilHall: @stratigos I didn't know about the `with` control. cool!

WilHall:
@tysongach 

> Should we display a flash message, too?

I think the inline message is enough, personally. For flash messages could be a good addition in the future, but we would need to write a facility for them.

> Curious what the output of `<%= error_tag f, :interests %>` is…is it associated with the input in some way?

It is just a span. Regarding your other comment about screen readers, I have updated the input to be described by the error text and it appears to read well now with VoiceOver. Unfortunately, due to the way the form gets rendered it is not currently possible to add the error message into the label element which is the other way I can think to do this. 🙁

Issue #804, opened 2019-11-06
Showing multiple entries for multiple comments in a single thread


ThoughtbotThom:
To improve scroll-ability, would love if there was only 1 entry for every post one commented on. Since we don't see the comment in this view, showing multiple comments to the same post is redundant.

<img width="565" alt="Screen Shot 2019-11-06 at 2 07 37 PM" src="https://user-images.githubusercontent.com/6619621/68329438-32c82900-009f-11ea-80dc-da5eb14b3be3.png">

germsvel:
@ThoughtbotThom are you seeing this on that new "Your comments" tab? (weird name for it too). Just want to make sure I know how we can reproduce that error.

ThoughtbotThom:
@germsvel sorry, yes this is in the "Your Comments" tab

PR #825, opened 2019-12-30
Show one announcement per comment in "Your Comments"

germsvel:
Closes https://github.com/thoughtbot/constable/issues/804

What changed?
=============

Prior to this change, if someone commented multiple times on the same announcement, clicking on "Your Comments" would show them multiple entries for the same announcement (one entry per comment).

We now include a DISTINCT on the query to only show one announcement in "Your Comments" tab regardless of how many times the user has commented on it.

### DIFF
--- a/test/models/announcement_test.exs
+++ b/test/models/announcement_test.exs
@@ -129,5 +129,19 @@ defmodule Constable.AnnouncementTest do
       assert announcement_with_user_comment.id in announcement_ids
       refute announcement_without_user_comment.id in announcement_ids
     end
+
+    test "includes each announcement once, even with multiple comments by user" do
+      user = insert(:user)
+      announcement = insert(:announcement)
+
+      insert_pair(:comment, announcement: announcement, user: user)
+
+      found_announcement =
+        user.id
+        |> Announcement.with_comments_by_user()
+        |> Repo.one()
+
+      assert found_announcement.id == announcement.id
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


test/models/announcement_test.exs
```diff
@@ -129,5 +129,19 @@ defmodule Constable.AnnouncementTest do
...
+    test "includes each announcement once, even with multiple comments by user" do
+      user = insert(:user)
+      announcement = insert(:announcement)
+
+      insert_pair(:comment, announcement: announcement, user: user)
+
+      found_announcement =
+        user.id
+        |> Announcement.with_comments_by_user()
+        |> Repo.one()
```
stratigos: It has been a while since Ive worked with Ecto, but doesnt `Repo.one` guarantee only one record is returned? Would this better protect against a false positive if you did a `Repo.all` and then asserted that only one record was returned in the result?

If not, could you show me where my understanding of `Repo.one` is off, or my understanding of this spec is off? Thanks :)

germsvel: Yeah, I always find `Repo.one` a bit confusing. 

`Repo.one` will _raise_ if more than one value is being returned. So it's not doing something like a `limit 1` (that would be [Repo.first/1](https://hexdocs.pm/ecto/Ecto.Query.html#first/2)). Instead, it's meant as a function that guarantees one or none to be returned. 

From [the Repo.one docs](https://hexdocs.pm/ecto/Ecto.Repo.html#c:one/2): 

> Fetches a single result from the query. Returns nil if no result was found. Raises if more than one entry.

The [`!` version of it](https://hexdocs.pm/ecto/Ecto.Repo.html#c:one!/2) will additionally raise if no records are found. 

Does that make more sense?

stratigos: Ah, my eyes glazed right over that `raises if...` part! TYSM for highlighting that. This makes total sense now, and is a nice guarantee that the test always produces a true positive or negative.

PR #849, opened 2020-02-24
Format codebase with mix format

germsvel:
Elixir's formatter was introduced back in version 1.6 ([read the announcement](https://elixir-lang.org/blog/2018/01/17/elixir-v1-6-0-released/)).

We use the formatter in other projects like [ExMachina](https://github.com/thoughtbot/ex_machina/), and it is very beneficial for standardizing the style of the project. We also seem to have added the `.formatter.exs` file in commit 7e1252ea so it seems like we already intended to use it.

Why format everything in a single commit?
-----------------------------------------

From my experience introducing formatting into other projects, it's easiest to run it once across the entire codebase and add a check in CI that ensures we no longer introduce code that is not formatted (we do that in this commit). It has the downside of ruining commit history a bit when using `git blame`, but seems to be the fastest way to introduce a formatter into a codebase.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -69,6 +69,8 @@ jobs:
           key: v4-npm-cache
           paths: "assets/node_modules"
 
+      - run: mix format --check-formatted
+
       - run:
           name: Wait for DB
           command: dockerize -wait tcp://localhost:5432 -timeout 1m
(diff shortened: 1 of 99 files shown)


.circleci/config.yml
```diff
@@ -69,6 +69,8 @@ jobs:
           key: v4-npm-cache
           paths: "assets/node_modules"
 
+      - run: mix format --check-formatted
```
germsvel: This ensures we don't introduce changes that haven't been formatted into the codebase in the future

PR #855, opened 2020-03-10
add instructions to readme for local env and data setup

lady3bean:
This PR adds instructions for adding environment variables and data from the staging server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ Run:
   ```
 
 ### Set up local data and env vars for Google OAuth
+
 You'll want to have certain environment variables and user data set up so you
 can log into the app locally with your thoughtbot email.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -50,6 +50,19 @@ Run:
   $ bin/setup
   ```
 
+
+### Add Local Environment Variables
+Run `cp .sample.env .env` to create a new .env file
```
germsvel: The `bin/setup` script [already does this](https://github.com/thoughtbot/constable/blob/master/bin/setup#L6-L10), so I'm not sure if it makes sense to call out explicitly. What do you think about removing this first line?

README.md
```diff
@@ -50,6 +50,19 @@ Run:
...
+In `.env`, there are three values from the sample file that say get-from-staging,
+CLIENT_ID, CLIENT_SECRET, and HUB_API_TOKEN. Replace these with the values from
+staging. Find them by running `heroku config -r staging`.
+
+
+### Copy the staging database to your development database
+Run `heroku pg:pull STAGING_DB_ENV_VAR constable_api_development -r staging`,
+where STAGING_DB_ENV_VAR is found from heroku staging. This will copy the data
+over so you have example data to work with.
```
germsvel: What do you think about adding some context as to why you would want to copy some of the database data and set `CLIENT_ID` and `CLIENT_SECRET`? (Namely that they are all needed in order to make google auth work locally)

Now that I'm writing this, I wonder if it makes sense to combine these two new sections into a single section that is titled something like `### Setting up data for google auth`. Not sure. What are your thoughts?

README.md
lady3bean: sure!

README.md
lady3bean: I like it! Will update.

PR #856, opened 2020-03-10
avoid module attributes to avoid compile time vs runtime bug

lady3bean:
This PR replaces the use of a module attribute for the permitted email domain when creating a user. 

### Why?
The issue with the module attribute is that is is evaluated at compile time, not runtime (https://ropig.com/blog/be-careful-when-using-elixirs-module-attributes/). This can cause a bug where the value is not there, because `System.get_env` can only be run at runtime. I ran into this today when trying to log in, and I've run into it before in phoenix projects.

### DIFF
--- a/lib/constable_web/controllers/auth_controller.ex
+++ b/lib/constable_web/controllers/auth_controller.ex
@@ -10,7 +10,8 @@ defmodule ConstableWeb.AuthController do
     UserInterest
   }
 
-  @permitted_email_domain Application.fetch_env!(:constable, :permitted_email_domain)
+  defp permitted_email_domain, do: Application.fetch_env!(:constable, :permitted_email_domain)
+
   @one_year_in_seconds 365 * 24 * 60 * 60
 
   def index(conn, %{"browser" => "true"}) do
@@ -36,7 +37,7 @@ defmodule ConstableWeb.AuthController do
     case find_or_insert_user(email, name) do
       nil ->
         conn
-        |> put_flash(:error, "You must sign up with a #{@permitted_email_domain} email address")
+        |> put_flash(:error, "You must sign up with a #{permitted_email_domain()} email address")
         |> redirect(external: "/")
 
       user ->
@@ -87,7 +88,7 @@ defmodule ConstableWeb.AuthController do
       nil ->
         conn
         |> put_status(403)
-        |> json(%{error: "must sign up with a #{@permitted_email_domain} email"})
+        |> json(%{error: "must sign up with a #{permitted_email_domain()} email"})
 
       user ->
         conn
@@ -122,7 +123,7 @@ defmodule ConstableWeb.AuthController do
 
       {:error, _changeset} ->
         Logger.info(
-          "Email address `#{email}` not from permitted `#{@permitted_email_domain}` domain"
+          "Email address `#{email}` not from permitted `#{permitted_email_domain()}` domain"
         )
 
         nil
--- a/test/controllers/auth_controller_test.exs
+++ b/test/controllers/auth_controller_test.exs
@@ -5,9 +5,8 @@ defmodule ConstableWeb.AuthControllerTest do
   require Constable.Pact
 
   @google_authorize_url "https://accounts.google.com/o/oauth2/auth"
-  @permitted_email_domain Application.fetch_env!(:constable, :permitted_email_domain)
-
-  def valid_email_address, do: "fake@#{@permitted_email_domain}"
+  def valid_email_address, do: "fake@#{permitted_email_domain()}"
+  defp permitted_email_domain, do: Application.fetch_env!(:constable, :permitted_email_domain)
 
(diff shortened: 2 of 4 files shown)


lib/constable/models/user.ex
```diff
@@ -71,9 +71,10 @@ defmodule Constable.User do
   defp require_permitted_email_domain(changeset) do
     changeset
     |> validate_change(:email, fn :email, value ->
-      case String.split(value, "@") do
-        [_, @permitted_email_domain] -> []
-        _ -> [email: "must be a member of #{@permitted_email_domain}"]
+      if String.split(value, "@") == permitted_email_domain() do
```
germsvel: Is this change equivalent to what we had before? And is it working in this current form?  

It seems to me like this equality will never evaluate to `true` because `String.split/2` will return a list of elements, but `permitted_email_domain()` is a string, right? So for example: 

```elixir
String.split("random@example.com", "@")
# => ["random", "example.com"]

String.split("someotherstring", "@")
# => ["someotherstring"]
```

In both of the cases above, you'd have: 

```elixir
## first example
if ["random", "example.com"] == "thoughtbot.com" do

## second example
if ["someotherstring"] == "thoughtbot.com"
```

Can you confirm that this is working as expected?

lib/constable_web/controllers/auth_controller.ex
```diff
@@ -10,7 +10,8 @@ defmodule ConstableWeb.AuthController do
     UserInterest
   }
 
-  @permitted_email_domain Application.fetch_env!(:constable, :permitted_email_domain)
+  def permitted_email_domain, do: Application.fetch_env!(:constable, :permitted_email_domain)
```
germsvel: What do you think about making these functions private? That way, we don't expose them outside of this module, which keeps the behavior in-line with how the module attribute was working.

lib/constable/models/user.ex
```diff
@@ -4,14 +4,14 @@ defmodule Constable.User do
...
 
-  @permitted_email_domain Application.fetch_env!(:constable, :permitted_email_domain)
-
   defimpl Bamboo.Formatter do
     def format_email_address(user, _opts) do
       {user.name, user.email}
     end
   end
 
+  def permitted_email_domain, do: Application.fetch_env!(:constable, :permitted_email_domain)
```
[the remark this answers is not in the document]
lady3bean: yeah, the environment variable is available at compilation time, so I think the issue is that the value of the variable may or not be there during compilation time. In production, you don't want to use module attributes with ENV variables because those variables might not be available during compilation. But in my experience, this problem is wonky, like, it sometimes works anyway, but in different environments, it may not.

lib/constable/models/user.ex
lady3bean: it must not have been working as expected, though I did test it out! I'm gonna push up a change for this, which I've tested more thoroughly.

lib/constable/models/user.ex
```diff
@@ -72,9 +72,15 @@ defmodule Constable.User do
     changeset
     |> validate_change(:email, fn :email, value ->
       case String.split(value, "@") do
-        [_, @permitted_email_domain] -> []
-        _ -> [email: "must be a member of #{@permitted_email_domain}"]
-      end
+        [_, domain] ->
```
lady3bean: you can't use a function in the pattern match here, which is why I was reaching for something other than the original line

test/controllers/auth_controller_test.exs
```diff
@@ -5,9 +5,8 @@ defmodule ConstableWeb.AuthControllerTest do
   require Constable.Pact
 
   @google_authorize_url "https://accounts.google.com/o/oauth2/auth"
-  @permitted_email_domain Application.fetch_env!(:constable, :permitted_email_domain)
-
-  def valid_email_address, do: "fake@#{@permitted_email_domain}"
+  def valid_email_address, do: "fake@#{permitted_email_domain()}"
```
lady3bean: I didn't use a private function here, cause it looks like it's used like it must be public - in `AuthControllerTest.valid_email_address()`, a few lines down

lady3bean:
is CirlceCI using a different `mix format` than me, perhaps? When I run `mix format`, it says there's nothing to format, but this PR is failing on Circle.

germsvel:
@lady3bean I don't see `mix format` failing in the [latest build](https://circleci.com/gh/thoughtbot/constable/2884). I think a couple of tests are failing though

lib/constable/models/user.ex
germsvel: What do you think about using [`String.ends_with?/2`](https://hexdocs.pm/elixir/String.html#ends_with?/2) here instead of doing an `if` statement within the `case` statement? 

```elixir
if String.ends_with?(value, "@#{permitted_email_domain()}") do
  []
else
  [email: "must be a member of #{permitted_email_domain()}"]
end
```

PR #881, opened 2020-05-14
Add pagination to announcements API

sharplet:
The status quo for /api/announcements is to download the entire history
of announcements. There are numerous problems with this.

Introduce a simple cursor-based pagination scheme:

- The default page size is 50, and can be customised with the `page_size` query parameter.
- With no pagination parameters, the most recent 50 announcements are returned.
- Passing `after=:announcement_id` fetches the next 50 most recent announcements after the one identified.
- Passing `before=:announcement_id` traverses the announcements in the opposite direction.
- Currently no metadata about the whole collection is returned, so in order to find the end of the list a client must keep requesting announcements until the API returns less than the `page_size`.

Co-authored-by: Dan Barber <dan@danbarber.me>

### DIFF
--- a/lib/constable_web/controllers/api/announcement_controller.ex
+++ b/lib/constable_web/controllers/api/announcement_controller.ex
@@ -8,8 +8,44 @@ defmodule ConstableWeb.Api.AnnouncementController do
 
   plug(Constable.Plugs.Deslugifier, slugified_key: "id")
 
-  def index(conn, _params) do
-    announcements = Repo.all(Announcement)
+  def index(conn, params = %{"after" => after_id}) do
+    limit = Map.get(params, "page_size", 50)
+    cursor = Repo.get(Announcement, after_id)
+
+    announcements =
+      Announcement
+      |> where([a], a.inserted_at < ^cursor.inserted_at)
+      |> order_by(desc: :inserted_at)
+      |> limit(^limit)
+      |> Repo.all()
+
+    render(conn, "index.json", announcements: announcements)
+  end
+
+  def index(conn, params = %{"before" => before_id}) do
+    limit = Map.get(params, "page_size", 50)
+    cursor = Repo.get(Announcement, before_id)
+
+    announcements =
+      Announcement
+      |> where([a], ^cursor.inserted_at < a.inserted_at)
+      |> order_by(:inserted_at)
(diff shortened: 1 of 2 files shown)


lib/constable_web/controllers/api/announcement_controller.ex
```diff
@@ -8,8 +8,44 @@ defmodule ConstableWeb.Api.AnnouncementController do
...
+    limit = Map.get(params, "page_size", 50)
+    cursor = Repo.get(Announcement, before_id)
+
+    announcements =
+      Announcement
+      |> where([a], ^cursor.inserted_at < a.inserted_at)
+      |> order_by(:inserted_at)
+      |> limit(^limit)
+      |> Repo.all()
+      |> Enum.reverse()
```
germsvel, line 35: Curious, why are we doing an `Enum.reverse()` here instead of sorting the articles in sql?

lib/constable_web/controllers/api/announcement_controller.ex
```diff
@@ -8,8 +8,44 @@ defmodule ConstableWeb.Api.AnnouncementController do
 
   plug(Constable.Plugs.Deslugifier, slugified_key: "id")
 
-  def index(conn, _params) do
-    announcements = Repo.all(Announcement)
+  def index(conn, params = %{"after" => after_id}) do
```
germsvel, line 11: What do you think about adding some tests for each of these scenarios? I'm not sure what the state of the api tests are, but maybe a good idea?

sharplet: I think @danbee mentioned he was planning to add some tests.

lib/constable_web/controllers/api/announcement_controller.ex
sharplet: AFAIK, a `LIMIT` clause doesn't work in the reverse direction in SQL. So in order to get the 50 articles _behind_ a given article, we came up with this method:

1. Reverse the sort order in SQL and apply the limit.
2. Fetch the records into memory and reverse them again to return to the original order (descending by publish time).

Does that make sense?

germsvel: 👍 that does make sense. Thanks for the explanation!

PR #888, opened 2020-05-26
Remove notion of models

germsvel:
What changed?
==============

Back in 7d37d54a25855e4ef459a501dc0879319bbffe88 we made most of the directory structure changes needed to conform to Phoenix 1.3 apps. One of the things we never got around to doing, however, was removing the concept of `models`in favor of simply having modules that have Ecto schemas. [This comment](https://github.com/thoughtbot/constable/issues/450#issuecomment-466423309) hinted at the possibility of returning to do this someday. This commit does just that.

We move `lib/constable/models` out of the models directory. Note that we only need to rename the files, not the module names themselves. The namespacing of modules already assumed the correct file structure: eg. the `Constable.Announcement` module is found in `lib/constable/announcement.ex`. It never had a `Model` namespace.

We also change the `use ConstableWeb, :model` to `use ConstableWeb, :schema` so that we can continue to include common Ecto schema functions in those modules that will have database interactions. Finally, we update the `ModelCase` helper to be a `DataCase` helper to indicate that it's a test helper that interacts with database data.


test/support/data_case.ex
```diff
@@ -1,4 +1,4 @@
-defmodule Constable.ModelCase do
+defmodule Constable.DataCase do
```
WilHall, line 1: Small naming question - if we're renaming `model` to `schema` in other places, would `SchemaCase` make sense here instead of `DataCase`?

germsvel: That's a good question. New Phoenix apps generate the `DataCase` for this, which is why I went this way. This is the template: https://github.com/phoenixframework/phoenix/blob/master/installer/templates/phx_ecto/data_case.ex

PR #890, opened 2020-05-29
Switch to Wallaby.Chrome and update deps

germsvel:
Commit 0acffe0 bumped Wallaby's version from 0.24.1 to 0.25.0.
Unfortunately, that broke the build because Wallaby's experimental
Chrome was moved out of experimental phase and into the default.

The fix is a simple config change. But we also encountered an issue with
dependencies that could have been brought about by the upgrade in
Wallaby or bumping phoenix_live_reload from 1.2.1 to 1.2.2 (see
24ceb82). So we have to do some cleaning of dependencies and update a
couple. The particular issue we were running into was that of mix config
not being able to determine the correct version of the `jason`
dependency for tests.


germsvel:
@mjankowski I'm going to merge this because it's blocking other work, but I wanted to ping you just in case you wanted to know about it. 😄 I hope that's alright.

Issue #887, opened 2020-05-26
Announcements page is very slow to load


sharplet:
I haven't measured this very precisely, but I'm noticing that when I visit `/announcements` the page currently takes well over 30 seconds to load.

germsvel:
I did a little bit of digging into this. I believe the issue is coming from fetching the profile urls and image urls from hub. We used to use gravatar, and that just required generating a url hash for each avatar. But now that we use hub, we try to memoize the urls. But when the cache isn't warm, this can take a really long time. I'll try to get to this later this week, but if I can't, I hope this note helps someone else fix this.

germsvel:
I'm working on a solution for now where we no longer memorize those values, but instead, we fetch them from hub every so often and store them in the database. That way, when we load a page, we already have the values. 

This is still a little weird because one of the best times to do that is when the user is created. But the values for creation are (I think) coming from hub. So we have something like this going on, 

```
hub | -- create a user (email, name)       --> | constable 
hub | <-- what is the user's profile info? --  | constable
hub | -- profile_url and profile_image_url --> | constable
```

So ultimately, the best solution would be for hub to just provide this information to us whenever a user is created (currently hub only sends the email and name). Until then, however, I think fetching the info whenever we want/need it would work fine.

PR #891, opened 2020-05-29
Fetch Profiles info rather than memoize calls

germsvel:
Fixes https://github.com/thoughtbot/constable/issues/887

What?
=====

We make a change in how and when we fetch profile information for users (profile_url and profile_image_url). Currently in production we memoize the calls to hub when we're trying to load a user's profile image url or profile image. The problem with that is that if the cache isn't warm, we have to make N queries to hub trying to fetch every user's image url when loading a list of announcements, where N is the number of users that have made a comment, made an announcement, plug the current user. That makes for a very slow page load, which sometimes times out.

In order to get around that, we now store the profile_url and profile_image_url as part of the users table. This gets rid of the N queries at load time, and we perform those actions when (a) a user is created, (b) a user is reactivated, or (c) daily via a mix task (the mix task only updates users who do not have a url set).

When a user is created or reactivated, we fetch the profile information asynchronously. To achieve this, we add a `Constable.TaskSupervisor` which can handle "one off" tasks.

The mix task, on the other hand, performs a "brute force" approach to fetching the image urls, making N requests for N users. But since we're doing it on a background job, this seems to be fine for now.

Finally, an even better approach that can be taken in the future (though it requires changes in hub) would be for hub to provide the information to us on user creation. I believe hub is already the source of truth for the email and name, and it provides those to constable on user creation (see api/user_controller#create). If it could also provide the profile_url and profile_image_url at that point, we wouldn't have to fetch them afterwards.

Currently, we roughly have this workflow on user creation:

```
hub | -- create a user (email, name)       --> | constable
hub | <-- what is the user's profile info? --  | constable
hub | -- profile_url and profile_image_url --> | constable
```

If hub could provide the information needed upon creation, we would simplify the workflow to:

```
hub | -- create a user (email, name, profile fields) --> | constable
```

### DIFF
--- a/lib/constable/serializers/user.ex
+++ b/lib/constable/serializers/user.ex
@@ -9,7 +9,7 @@ defimpl Poison.Encoder, for: Constable.User do
       id: user.id,
       email: user.email,
       name: user.name,
-      profile_image_url: profile_provider().image_url(user),
+      profile_image_url: user.profile_image_url,
       user_interests: user.user_interests,
       subscriptions: user.subscriptions
     }
--- a/lib/constable/services/hub_profile_provider.ex
+++ b/lib/constable/services/hub_profile_provider.ex
@@ -1,24 +1,20 @@
 defmodule Constable.Services.HubProfileProvider do
-  alias ConstableWeb.Endpoint
   alias Constable.Services.ProfileProvider
   @behaviour ProfileProvider
-  use Memoize
-
-  @default_image_path "/images/ralph.png"
 
   @impl ProfileProvider
-  defmemo profile_url(user) do
+  def fetch_profile_url(user) do
     "#{Application.fetch_env!(:constable, :hub_url)}/people/#{fetch_slug(user)}"
   end
 
   @impl ProfileProvider
-  defmemo image_url(user) do
+  def fetch_image_url(user) do
     case Neuron.query(person_image_url_by_email_query(user.email)) do
       {:ok, %Neuron.Response{body: %{"data" => %{"person" => %{"image_url" => image_url}}}}} ->
         image_url
 
       _ ->
-        "#{Endpoint.url()}#{@default_image_path}"
+        nil
     end
   end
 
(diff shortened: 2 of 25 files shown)


lib/constable/services/hub_profile_provider.ex
```diff
@@ -1,24 +1,20 @@
...
 
   @impl ProfileProvider
-  defmemo profile_url(user) do
+  def fetch_profile_url(user) do
     "#{Application.fetch_env!(:constable, :hub_url)}/people/#{fetch_slug(user)}"
   end
 
   @impl ProfileProvider
-  defmemo image_url(user) do
+  def fetch_image_url(user) do
```
germsvel, line 11: I updated these to be `fetch_` to disambiguate from the attributes in the user (e.g. `user.profile_url`).

mix.exs
```diff
@@ -56,7 +56,6 @@ defmodule Constable.MixProject do
       {:html_sanitize_ex, "~> 1.4.0"},
       {:httpoison, "~> 1.3", override: true},
       {:jason, "~> 1.1"},
-      {:memoize, "~> 1.2"},
```
WilHall, line 59: 🎉

priv/repo/migrations/20200527201059_add_profile_fields_to_users.exs
```diff
@@ -0,0 +1,10 @@
+defmodule Constable.Repo.Migrations.AddProfileFieldsToUsers do
+  use Ecto.Migration
+
+  def change do
+    alter table(:users) do
+      add :profile_url, :string
```
WilHall, line 6: I ❤️ having these directly on the user. I think this approach is a lot better!

lib/constable/serializers/user.ex
```diff
@@ -9,7 +9,7 @@ defimpl Poison.Encoder, for: Constable.User do
       id: user.id,
       email: user.email,
       name: user.name,
-      profile_image_url: profile_provider().image_url(user),
+      profile_image_url: user.profile_image_url,
```
WilHall, line 12: Accessing this through the user is a lot clearer!

germsvel:
Thanks for the review @WilHall!

Issue #880, opened 2020-05-14
Plug.Conn is imported but not used in router.ex


sushant12:
https://github.com/thoughtbot/constable/blob/master/lib/constable_web.ex#L61

I believe that we should remove it if it's not  used.

germsvel:
@sushant12 thanks for opening this issue. Importing `Plug.Conn` in the router is the standard way Phoenix generates apps (you can see that [in the Phoenix template](https://github.com/phoenixframework/phoenix/blob/ea10baae82a9b849b9b4b4c1977dcfc344662863/installer/templates/phx_single/lib/app_name_web.ex#L66) for generating a new phoenix app), so I'm going to close this issue along with the associated PR.

PR #903, opened 2020-06-24
remove import of plug.conn in router

lady3bean:
Closes https://github.com/thoughtbot/constable/issues/880

### DIFF
--- a/lib/constable_web.ex
+++ b/lib/constable_web.ex
@@ -58,7 +58,6 @@ defmodule ConstableWeb do
     quote do
       use Phoenix.Router
       use Honeybadger.Plug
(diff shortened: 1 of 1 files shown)


germsvel:
@lady3bean I don't think we should make this change. Importing `Plug.Conn` in the router is the [standard way Phoenix generates apps](https://github.com/phoenixframework/phoenix/blob/ea10baae82a9b849b9b4b4c1977dcfc344662863/installer/templates/phx_single/lib/app_name_web.ex#L66). Do you mind if we close this?

brimatteng:
No one is still actively working on this

PR #911, opened 2020-07-06
Print Chromedriver, Elixir/Erlang, node/npm versions in CI

germsvel:
What? 
=====

I noticed the last two builds in CI have failed (https://circleci.com/gh/thoughtbot/constable/2989 and https://circleci.com/gh/thoughtbot/constable/2987). For some reason, I cannot get the builds to fail in this PR, but I still think the work in this PR makes sense to commit. 

When CI fails, I frequently want to see which versions of dependencies we are running. We now print the versions of Chromedriver, Elixir, Erlang, npm, and node.js so when builds fail we can quickly check which versions are running in CI and if a discrepancy in versions contributes to the issues at hand.

## Screenshots

Now we get these nice sections in CI

Screen Shot 2020-07-06 at 12 27 03 PM

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,6 +22,18 @@ jobs:
     steps:
       - checkout
 
+      - run:
+          name: Chromedriver version
+          command: chromedriver -v
+
+      - run:
+          name: Elixir & Erlang versions
+          command: elixir -v
+
+      - run:
+          name: npm & node versions
+          command: npm -v && node -v
+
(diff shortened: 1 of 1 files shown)


lady3bean:
This will be helpful. I had noticed that a newer version of earmark must have been being used by CI that was causing the build to fail, and made a PR to limit the version here - https://github.com/thoughtbot/constable/pull/904. Had toy  dig to figure out that was what was happening, so definitely +1 to this change.

Issue #909, opened 2020-07-02
Constable is unusably slow for some users


georgebrock:
@eunjeeyoon and @sharplet, and maybe others, have experienced very slow load times.

Here's a screenshot from Eunjee, which shows an initial page load of 0.2s but a completion time of 1.2 minutes.

More discussion in Slack: https://thoughtbot.slack.com/archives/C03HS40ML/p1593725041000500

<img width="554" alt="Screen shot of Chrome network activity on a slow load" src="https://user-images.githubusercontent.com/35861/86411521-0777f700-bc8b-11ea-899b-6c484546d6c0.png">

sharplet:
Here's a snapshot from Safari's network debugger:
<img width="1124" alt="Screen Shot 2020-07-02 at 5 42 19 PM" src="https://user-images.githubusercontent.com/147686/86411850-bddbdc00-bc8b-11ea-87f4-62007c1ec8c8.png">

sharplet:
Note the following metrics: Duration 0ms, **Latency 110.14s**. That shubox URL never loads, and the page eventually pops in when it times out. When I disable JavaScript, the page loads incredibly quickly.

sharplet:
Possibly-related shubox issue: https://github.com/thoughtbot/constable/issues/866

sharplet:
Can we make it so this doesn't block page load?

georgebrock:
We could probably put an `async` or `defer` on this `<script>` element:

https://github.com/thoughtbot/constable/blob/master/lib/constable_web/templates/layout/app.html.eex#L22

We should probably also fix the broken script URL. I also see it failing to load, but for whatever reason it doesn't take so long to fail for me (in Safari or Chrome)

sharplet:
I get the impression from [Shubox's setup documentation for Constable](https://dashboard.shubox.io/domains/constable-io) that their JavaScript is meant to be embedded in our codebase. Perhaps it used to be hosted at shubox.io and that's no longer supported?

sharplet:
FWIW: the slowdown seems like a DNS issue. I'd been playing around with content blockers in Safari to try and disable JavaScript altogether for Constable, and I've noticed that the timeout has disappeared for me. Maybe my DNS cache knows that js.shubox.io is no longer valid?

mjankowski:
I think we are using an old/legacy method of including shubox. Might be worth review.

germsvel:
I'll try to handle this in two separate PRs.

[...]

PR #913, opened 2020-07-06
Update shubox to use new api

germsvel:
Closes https://github.com/thoughtbot/constable/issues/909

What changed?
=============

Constable was slow to load, seemingly in part because of shubox. So we move the loading of shubox to be async in a055a97231bc4b1a13e7b50c739ad0c1809082db.

While researching that, we noticed we have an old way of using shubox.
This commit updates shubox to the new way:

1. Install it via npm (no longer need to hit a specific url asynchronously)

2. Update the `s3urlTemplate` to `successTemplate` init option. The `s3urlTemplate` is deprecated. See https://github.com/shuboxio/shubox.js/blob/master/README.md#successtemplate

3. Update configuration files, sample.env, etc. to refer to `SHUBOX_KEY` and `shubox_key` instead of `shubox_script_url`. Shubox now has a unique key for each environment/app, so we need to use the correct ones. Note that there is a sandbox mode for development: so if desired, developers/designers can put the sandbox key in their .env files and upload images locally. All keys are available in shubox's dashboard.

There are some improvements we could make. In particular, we could add an `uploadingTemplate` to shubox to notify the user that we're in the process of uploading the image. But we found some difficulty with that and we leave that work for the future.

## Uploading images to new announcements and comments

uploading-image-announcement

uploading-image-comment


lib/constable_web/templates/layout/app.html.eex
```diff
@@ -18,8 +18,8 @@
     <link rel="apple-touch-icon" sizes="180x180" href="/apple-touch-icon-180x180.png" />
     <%= if current_user(@conn) do %>
       <script>window.userToken = "<%= @current_user.token %>";</script>
+      <script>window.shuboxKey = "<%= Application.fetch_env!(:constable, :shubox_key) %>";</script>
```
jutonz, line 21: This is sort of outside the scope of what this PR is aiming to do, but how do you feel about moving this JS into a JS file and just storing the token in HTML?

So for example, in the HTML:

```html
<head>
  <meta data-role="shubox-key" data-value="<%= Application.fetch_env!(:constable, :shubox_key) %>">
<head>
```

and then loading it in JS using something like this (untested code):

```js
$(document).onLoad(() => {
  const shuboxKey = $("[data-role=shubox-key]").attr("data-value");
  window.shuboxKey = shuboxKey; 
})
```

I think that'd be a better separation of concerns since we could keep javascript in JS files and markup in HTML files. What do you think?

germsvel: Yeah, this is a good point. To be honest, I never really know what's the best way to pass data from Elixir/Phoenix -> JavaScript. I have used data-attributes in the past, but sometimes find them more work than is needed, especially if the value is global. 

Are you okay if we merge this as is? I wouldn't mind creating an issues for Constable where we revisit how we're handling JavaScript. There's quite a bit of it in html templates. Not sure if we can find a better way to do that.

jutonz: Gotcha, that makes sense. I agree that it's not super clear what the best practice is. I made this suggestion after re-reading the [rails guides] (link in coffeescript) on the subject, but even if that's the route we want to go I agree it makes sense to tackle it all at once as part of a concentrated effort. Totally happy to merge this as-is! 👍

[rails guides]: https://guides.rubyonrails.org/working_with_javascript_in_rails.html#unobtrusive-javascript

Issue #915, opened 2020-07-06
Improve "Uploading..." file upload experience


germsvel:
When uploading a file, the user does not get any indication that the upload is in progress. It would be nice to add some type of `Uploading ...`. 

[Shubox](https://shubox.io/), which we use in Constable, seems to have an option for an [uploadingTemplate](https://github.com/shuboxio/shubox.js/blob/master/README.md#uploadingtemplate). In my brief attempt to use it, however, I discovered that the text inserted for the upload is not removed once the image is added. 

If we could find a way to make this work, it would be a nice improvement over the current workflow. 

### Visuals!

Here's a gif of me trying the uploading template. Note how the "Uploading ralph.png..." doesn't go away after the upload succeeds. 

![uploading-template-fail](https://user-images.githubusercontent.com/3245976/86635380-01428d00-bfa1-11ea-9f67-f0d1fdd289d6.gif)

PR #927, opened 2020-09-04
Add "Uploading ..." placeholder when uploading images

germsvel:
Closes https://github.com/thoughtbot/constable/issues/915

What changed?
============

We add an "Uploading ..." placeholder (it actually includes the filename) when uploading images into a constable post. Currently, when uploading an image, it is hard to know if the dropzone captured the image or not because the user gets no feedback about the upload being in progress. Shubox has an `uploadingTemplate` option but there was a bug that was only recently fixed in version `0.5.0`. So we update our Shubox version and make use of the `uploadingTemplate` option.


jutonz (approved):
Sweet, looks good!

PR #952, opened 2020-12-10
Update Phoenix to 1.5.x

germsvel:
What changed?
============

We update Phoenix's version to 1.5.x.

Changes made:

- Update dependencies
- Move PubSub into supervision tree
- Use `@inner_content` assign in layouts
- Change `use Phoenix.ConnTest` and `ChannelTest` -> `import`


.circleci/config.yml
```diff
@@ -39,46 +39,46 @@ jobs:
 
       - restore_cache:
           keys:
-            - v3-mix-cache-{{ .Branch }}-{{ checksum "mix.lock" }}
-            - v3-mix-cache-{{ .Branch }}
-            - v3-mix-cache
+            - v5-mix-cache-{{ .Branch }}-{{ checksum "mix.lock" }}
+            - v5-mix-cache-{{ .Branch }}
```
brimatteng, line 43: Is there something automated incrementing these or are they just done by hand?

brimatteng (approved):

.circleci/config.yml
germsvel: The `vX` are done by hand. The things between `{{ }}` are done by circle ci. So usually, the cache updates as we change dependencies (and it's per branch), but sometimes you need to do a hard break, so that's why we include the version.

PR #955, opened 2020-12-11
Use LiveView in interests index page

germsvel:
What changed?
============

We move the interests' index page to `InterestLive`. The current experience is a bit clunky: subscribing to an interest far down the page reloads the page causing it to go back up to the top. With LiveView we can handle that more smoothly.

In addition to moving the create/delete of user-interests to `InterestLive`, we also set the `user_id` in session (to get it from the live view). Doing so also revealed a little but in `FetchCurrentUser` plug that assumed we weren't setting a `current_user` assign with an inactive user, but it was actually setting a `nil` user. We now explicitly handle that.

Finally, LiveView does not play nicely with Turbolinks, since Tubolinks doesn't do a full page load and the `InterestLive` is mounted on the router. So, we do not use Turbolinks to go to the interest's index page since that prevents the correct loading of live view.

Co-authored-by: Mike Burns <mburns@thoughtbot.com>

<details><summary>Before LiveView</summary>

non-live-sub-unsub

</details>

<details><summary>With LiveView</summary>

live-sub-unsub

</details>


lib/constable/plugs/fetch_current_user.ex
```diff
@@ -10,8 +10,13 @@ defmodule Constable.Plugs.FetchCurrentUser do
       :not_signed_in ->
         conn
 
+      nil ->
```
mike-burns, line 13: Note for reviewers: this is to avoid an issue with `user.id` below.

lib/constable_web/templates/announcement/index.html.eex
```diff
@@ -27,7 +27,7 @@
 
     <%= if !@show_all do %>
       <%= link to: Routes.interest_path(@conn, :index),
-        data: [role: "view-all-interests"],
+        data: [role: "view-all-interests", turbolinks: false],
```
mike-burns, line 30: Note for reviewers: this is where we ran into issues with Turbolinks navigating to the interests page: it does not load the LiveView JS when that happens. We could work around that but for now this is fine.

mike-burns:
LGTM!

PR #959, opened 2020-12-28
Do not compile assets every `mix test`

germsvel:
What changed?
=============

Currently we compile our assets every time we run `mix test`. That's true regardless of whether we're running the entire test suite or a single test. That causes terribly slow feedback cycles when doing TDD.

In this commit, we remove the automatic compilation of assets when running `mix test`, and we create a new alias `mix assets.watch` to easily watch assets when we're making changes to CSS or JS.

Thus, if you're running tests iteratively and are making changes to CSS or JS, you can open a new pane and run `mix assets.watch` and do your normal TDD on a different pane.


lady3bean (approved):
Nice!

PR #1002, opened 2021-06-25
Press cmd+enter to add comment

jutonz:
When commenting on an announcement, you can now press cmd + enter to
create the comment.

To support Linux and Windows, you can also do ctrl + enter.

### DIFF
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -56,7 +56,17 @@ Hooks.CommentPreview = {
   },
 };
 let csrfToken = document.querySelector("meta[name='csrf-token']").getAttribute('content');
-let liveSocket = new LiveSocket('/live', Socket, {hooks: Hooks, params: {_csrf_token: csrfToken}});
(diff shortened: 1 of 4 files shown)


assets/js/app.js
```diff
@@ -56,7 +56,17 @@ Hooks.CommentPreview = {
...
+let liveSocket = new LiveSocket('/live', Socket, {
+  hooks: Hooks,
+  params: { _csrf_token: csrfToken },
+  metadata: {
+    keydown: (event, _element) => ({
+      key: event.key,
+      metaKey: event.metaKey,
+      ctrlKey: event.ctrlKey,
+    }),
+  },
```
jutonz, line 68: In order for info about modifier keys to be sent to the backend, we need to add it manually. Docs [here](https://hexdocs.pm/phoenix_live_view/bindings.html#key-events).

brimatteng (approved):
Looks rad!

PR #1020, opened 2021-07-23
Bump Elixir/Erlang Versions to 1.12.1/24.0

brimatteng:
Note: This is live on staging -> https://staging.constable.io/

## Why? 
- Earmark (used for Markdown parsing) and Poison (Used for JSON parsing) both have a updates which require more recent elixir versions. Upgrading makes it easier to keep dependencies up to date.
- New versions of elixir have tightened up the integration with Erlang for purposes of errors and logging and adding more clarity to messages. The elixir compiler has also had a performance tuning pass done promising faster build times. 

### Changes:
- Update the CircleCI image we use for testing
- Update Poison (This was throwing a ton of warnings when I upgraded elixir and I wanted to get it out of the way. Nothing we are doing with poison changed) 
- Run `mix format` to pass the format checker on circle
- Update `.tool-versions` and the `mix.exs` config


mix.exs
```diff
@@ -4,8 +4,8 @@ defmodule Constable.MixProject do
   def project do
     [
       app: :constable,
-      version: "0.1.0",
-      elixir: "~> 1.5",
+      version: "0.2.0",
```
brimatteng, line 7: Not sure what this version represents

